PR #155, opened 2025-12-23
Fix: Handle paths with spaces in parsePythonCommand

holgarsson:
## Problem

The `parsePythonCommand` function in `python-detector.ts` was splitting all Python paths by space:

```typescript
const parts = pythonPath.split(' ');
```

This breaks paths containing spaces, like macOS's default Electron userData path:
```
/Users/name/Library/Application Support/auto-claude-ui/python-venv/bin/python3
```

The path gets split into:
- command: `/Users/name/Library/Application`
- args: `['Support/auto-claude-ui/python-venv/bin/python3']`

This causes `ENOENT` errors when spawning Python processes.

## Solution

Only split by space for simple command names (like `py -3` on Windows). For absolute paths (starting with `/` or a drive letter), preserve the full path.

## Testing

Tested on macOS with the default `~/Library/Application Support/` path - the bug is now fixed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Python detection to correctly handle full-path invocations on Unix and Windows (including drive-letter paths and paths with spaces).
  * Enhanced script discovery so the packaged desktop app can locate bundled detector scripts in application resources.

* **Tests**
  * Added comprehensive unit tests validating Python command parsing, default selection, and discovery behavior.

### DIFF
--- a/auto-claude-ui/src/main/ipc-handlers/memory-handlers.ts
+++ b/auto-claude-ui/src/main/ipc-handlers/memory-handlers.ts
@@ -457,7 +457,9 @@ export function registerMemoryHandlers(): void {
         }
 
         // Find the ollama_model_detector.py script
+        // In packaged Electron apps, process.resourcesPath points to the Resources folder
         const possiblePaths = [
+          path.resolve(process.resourcesPath || '', 'auto-claude', 'ollama_model_detector.py'),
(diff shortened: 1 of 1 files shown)


AndyMik90:
This fix has been merged into PR #185 (fix/bug-fixes-2.7.2) along with PR #182 which addressed the same issue. Thank you for the contribution! 🙏

PR #187, opened 2025-12-23
fix: remove legacy path from auto-claude source detection

Jorisslagter:
## Changes

Fixes the spec runner path detection by removing the legacy auto-claude/runners path that no longer exists after the apps/ restructure.

### What changed
- Updated source path detection to only use valid paths: apps/backend and legacy root location
- Removed outdated auto-claude/runners path that was causing "Spec runner not found" errors

### Testing
- Verified spec runner detection works correctly with current repository structure
- Tested with settings.json configuration

Fixes the "Spec runner not found" error reported in the workspace review feature.


AndyMik90:
This fix has been merged into PR #185 (fix/bug-fixes-2.7.2). Thank you for the contribution! 🙏

Issue #232, opened 2025-12-24
Python venv creation fails with Python 3.9 - claude-agent-sdk requires Python 3.10+


tob-as:
## Description

When starting the Auto Claude UI for the first time, the Python environment initialization fails with the error:

```
Failed to run merge: spawn /Users/.../auto-claude/.venv/bin/python ENOENT
```

This occurs because the `claude-agent-sdk` package requires Python 3.10+, but the system may have Python 3.9 as the default, causing the virtual environment setup to fail.

## Root Cause

1. The `requirements.txt` specifies `claude-agent-sdk>=0.1.16`
2. This package requires Python 3.10 or higher
3. If the system's default `python3` is version 3.9.x, the venv creation succeeds but dependency installation fails
4. The Electron UI then tries to use a non-existent or incomplete Python environment

## Impact

- First-time setup fails silently
- Users see cryptic `ENOENT` errors when trying to use features like merge, QA, or autonomous builds
- The UI appears to work but backend operations fail

## Steps to Reproduce

1. Have Python 3.9.x as your system's default `python3`
2. Clone the Auto-Claude repository
3. Launch the Electron UI (`npm start` in `auto-claude-ui/`)
4. The UI starts but any backend operation fails with `ENOENT` error
5. Checking the venv reveals missing packages:
   ```bash
   cd auto-claude
   .venv/bin/pip list | grep claude-agent-sdk
   # Returns nothing - installation failed
   ```

### Option 1: Check Python Version in Setup Script (Recommended)

Add a Python version check in `auto-claude-ui/src/main/python-env-manager.ts`:

```typescript
private async validatePythonVersion(pythonCommand: string): Promise<boolean> {
  try {
    const result = await execAsync(\`\${pythonCommand} --version\`);
    const match = result.stdout.match(/Python (\d+)\.(\d+)/);
    if (match) {
      const major = parseInt(match[1]);
      const minor = parseInt(match[2]);
      return major === 3 && minor >= 10;
    }
  } catch (error) {
    return false;
  }
  return false;
}
```

Then use this before creating the venv and show a clear error message to users if their Python is too old.

### Option 2: Auto-detect Higher Python Version

Modify the Python detection logic to prefer Python 3.10+ when available:

```typescript
// In python-detector.ts or python-env-manager.ts
const pythonVersions = ['python3.13', 'python3.12', 'python3.11', 'python3.10', 'python3'];
for (const cmd of pythonVersions) {
[...]

Issue #215, opened 2025-12-23
Can't get the app to work


apvlv:
### Checklist

- [x] I searched existing issues and this hasn't been reported

### Area

Frontend

### Operating System

macOS

### Version

2.7.1

### What happened?

Thank you for the wonderful idea and the work on that project. However i couldn't get a single thing to work so far.. It doesn't find my gh installation and on other places it just says spawn python ENOENT error, even though I specified the path to python...

Are this bugs known to you?

### Steps to reproduce

Try to connect to github.
or
Analyze context

### Expected behavior

to function properly

### Logs / Screenshots

```shell

```

apvlv:
Whether automatic or manual token based auth to my claude max plan also doesn‘t work

<img width="480" height="186" alt="Image" src="https://github.com/user-attachments/assets/d239784c-9983-4d2c-a235-099f52992279" />

chuckjewell:
This was my journey as well. 

- Claude auth worked. 
- ollama detection did not work (installed via brew)
- github detection didnt work (installed via brew)
- Trying to generate a context: Failed to load project index - spawn python ENOENT (I do some python dev so all normal tools and versions installed)

This is on a pretty clean M4 max (15.6) set up with all normal dev tools in the terminal.

Past1che:
saving the same issues

<img width="822" height="335" alt="Image" src="https://github.com/user-attachments/assets/307acbd4-e41e-43eb-bde7-f54c5f18dd9f" />

lazyants:
> This was my journey as well.
> 
> * Claude auth worked.
> * ollama detection did not work (installed via brew)
> * github detection didnt work (installed via brew)
> * Trying to generate a context: Failed to load project index - spawn python ENOENT (I do some python dev so all normal tools and versions installed)
> 
> This is on a pretty clean M4 max (15.6) set up with all normal dev tools in the terminal.

Also on my Mac, gh and ollama installed via Homebrew and were not found

r3d91ll:
## Fix Confirmed: spawn python ENOENT Resolved

  I've successfully identified and fixed this issue on **Ubuntu 24.04**. The fix should also resolve the macOS reports.

  ### Root Cause

  The Python detector (`findPythonCommand()`) correctly finds Python at:
  - **macOS**: `/opt/homebrew/bin/python3` (Homebrew)
  - **Ubuntu**: `/usr/bin/python3` (system)

[...]

PR #241, opened 2025-12-24
fix: resolve Python detection and backend packaging issues

Furansujin:
## Summary
Fixes critical issues preventing project index loading on macOS:
- Backend files not found (wrong packaging path)
- Python compatibility errors (missing future annotations)
- Wrong Python interpreter used (system instead of Homebrew)

## Changes
1. **package.json**: Changed packaging destination from `auto-claude` to `backend` to match path-resolver.ts expectations
2. **config_parser.py**: Added `from __future__ import annotations` for Python 3.9+ compatibility with `dict | None` syntax
3. **project-context-handlers.ts**: Use `findPythonCommand()` to auto-detect correct Python instead of hardcoded `'python3'`
4. **python-detector.ts**: Prioritize Homebrew Python paths (`/opt/homebrew/bin/python3`) over system Python on macOS

## Root Cause
The app was looking for the backend in `Resources/backend/` but packaging put it in `Resources/auto-claude/`. Even when found, Python 3.9 would fail on union type syntax, and macOS would use the old system Python (3.9) instead of Homebrew Python (3.10+).

## Testing
- ✅ Tested on macOS Apple Silicon with packaged app
- ✅ Project index loads without errors
- ✅ Uses Homebrew Python 3.12 instead of system Python 3.9
- ✅ `analyzer.py` found at correct path
- ✅ No TypeError with `dict | None` syntax

## Screenshots
Before: "Failed to load project index - analyzer.py not found"
After: Project index loads successfully

Resolves issues reported in Discord/community.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced Python environment detection to dynamically resolve available installations instead of relying on static defaults
  * Improved compatibility with different Python installation methods on macOS
  * Maintained fallback functionality for Python command resolution across platforms

### DIFF
--- a/apps/backend/project/config_parser.py
+++ b/apps/backend/project/config_parser.py
@@ -6,8 +6,6 @@ Utilities for reading and parsing project configuration files
 (package.json, pyproject.toml, composer.json, etc.).
 """
 
-from __future__ import annotations
-
 import json
 import sys
 from pathlib import Path
--- a/apps/frontend/src/main/python-detector.ts
+++ b/apps/frontend/src/main/python-detector.ts
@@ -2,6 +2,27 @@ import { execSync } from 'child_process';
 import { existsSync } from 'fs';
 
 /**
+ * Find the first existing Homebrew Python installation.
+ * Checks common Homebrew paths for Python 3.
+ *
+ * @returns The path to Homebrew Python, or null if not found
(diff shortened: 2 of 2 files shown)


Furansujin:
fixe issue #232 and #215

PR #311, opened 2025-12-26
fix(python): use venv Python for all services to fix dotenv errors

AlexMadera:
Fix Python dotenv Errors on Windows

  Problem

  Services were spawning Python using findPythonCommand() which returns the bundled Python directly. However, dependencies like python-dotenv are only installed in the venv created from the bundled Python, causing ModuleNotFoundError: No module named 'dotenv' errors.

  Solution

  Created getConfiguredPythonPath() helper that returns the venv Python (with dependencies) when ready, and updated all services to use it.

## Summary by CodeRabbit

* **Chores**
  * Added a frontend verification script to validate Python bundling, resource configuration, venv capability, and required dependencies before packaging.
  * Centralized Python executable resolution across the frontend to a configurable provider, with consistent fallback semantics and updated services to use the configured path.

### DIFF
--- a/apps/frontend/scripts/verify-python-bundling.cjs
+++ b/apps/frontend/scripts/verify-python-bundling.cjs
@@ -60,7 +60,6 @@ if (pythonResource) {
 
 // Check 3: Test venv creation simulation
 console.log('\n3. Checking venv creation capability...');
-const testVenvPath = path.join(FRONTEND_DIR, '.test-venv-check');
 try {
   // Find system Python for testing
   let pythonCmd = process.platform === 'win32' ? 'python' : 'python3';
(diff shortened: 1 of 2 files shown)


MikeeBuilds (approved):
LGTM! 🚀

This is exactly the fix we needed. The `getConfiguredPythonPath()` helper ensures all services use the venv Python (with dependencies installed) instead of the bundled Python directly.

Verified the pattern is applied consistently across:
- agent-process.ts
- title-generator.ts
- memory-service.ts
- worktree-handlers.ts
- memory-handlers.ts
- changelog-service.ts
- insights/config.ts
- project-context-handlers.ts

The verification script is a nice bonus for CI checks.

Tested locally - approved! 👍

PR #444, opened 2025-12-31
fix: infinite loop in useTaskDetail merge preview loading

abe238:
## Summary
- Fixed infinite loop in `useTaskDetail.ts` that caused repeated merge preview loading
- The loop was caused by the cache-clearing effect and auto-load effect fighting each other in React Strict Mode
- Added `hasLoadedPreviewRef` to track loaded state across renders
- Removed excessive console.warn statements cluttering the console

## Problem
When viewing a task in human_review status, the console would be flooded with:
```
[useTaskDetail] Cleared merge preview cache for task: ...
[useTaskDetail] Auto-loading merge preview for task: ...
[useTaskDetail] loadMergePreview called for task: ...
```

This was the same infinite loop pattern found in App.tsx (PR #442).

## Root Cause
1. Effect at line ~200 clears `mergePreview` to null when task changes
2. Effect at line ~240 sees `!mergePreview` condition and calls `loadMergePreview()`
3. `loadMergePreview` sets preview data
4. React Strict Mode re-runs effects, effect 1 clears it again
5. Loop continues forever

## Solution
Used a ref (`hasLoadedPreviewRef`) to track whether preview has been loaded for the current task:
- Cache-clearing effect only clears when switching to a *different* task
- Auto-load effect checks the ref instead of the state to determine if loading is needed
- The ref persists across Strict Mode double-invokes, breaking the loop

## Test plan
- [x] Verify no infinite loop when viewing tasks in human_review status
- [x] Verify merge preview still loads correctly on first view
- [x] Verify switching between tasks loads fresh preview data

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented repeated/recursive task preview loading so task details load once per task, improving reliability and responsiveness.
  * Reduced noisy console output during preview fetches for cleaner logs and smoother user experience.
  * Improved loading state handling to stop spinners promptly when previews finish or fail.

### DIFF
--- a/apps/frontend/src/renderer/components/task-detail/hooks/useTaskDetail.ts
+++ b/apps/frontend/src/renderer/components/task-detail/hooks/useTaskDetail.ts
@@ -198,13 +198,9 @@ export function useTaskDetail({ task }: UseTaskDetailOptions) {
   // Track if we've already loaded preview for this task to prevent infinite loops
   const hasLoadedPreviewRef = useRef<string | null>(null);
 
-  // Clear merge preview cache when task changes to ensure fresh data is fetched
-  // This invalidates any stale cached data (e.g., old uncommitted changes status)
+  // Clear merge preview state when switching to a different task
   useEffect(() => {
-    // Only clear if we're switching to a different task
     if (hasLoadedPreviewRef.current !== task.id) {
-      const storageKey = `mergePreview-${task.id}`;
-      sessionStorage.removeItem(storageKey);
       setMergePreview(null);
       hasLoadedPreviewRef.current = null;
     }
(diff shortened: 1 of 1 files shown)


abe238:
## Addressed code review feedback

Thanks @gemini-code-assist and @coderabbitai for the thorough reviews! 

**Fixed in commit 293e8ba:**

1. **Infinite retry loop on failures** (Critical) - Moved `hasLoadedPreviewRef.current = task.id` to the `finally` block so the ref is set regardless of success or failure. This prevents infinite retries when the API call fails.

2. **Removed dead sessionStorage code** (Minor) - Removed unused `sessionStorage.setItem/removeItem` calls since there was no corresponding `getItem` to restore the cached data.

The auto-load effect will now only attempt to load the preview once per task, and users can manually retry via UI if needed.

abe238:
I have read the CLA Document and I hereby sign the CLA

PR #452, opened 2025-12-31
fix(macOS): respect user-configured pythonPath setting for venv creation

webdevtodayjason:
## Problem
The `pythonPath` setting from the UI was being ignored when creating the Python virtual environment. This caused macOS users with Miniconda/pyenv to have their venv created with the system Python (3.9.6) instead of their configured Python (e.g., 3.13.2).

## Root Cause
`PythonEnvManager.findSystemPython()` was only checking for bundled/system Python via auto-detection, completely ignoring the user's `pythonPath` setting from the UI.

## Solution
- Add `configure(pythonPath)` method to `PythonEnvManager` to accept user's pythonPath
- Update `findSystemPython()` to check configured path **first** before auto-detection
- Call `pythonEnvManager.configure()` in `main/index.ts` with `settings.pythonPath`

## Priority Order for Python Detection (after fix)
1. User-configured `pythonPath` from settings
2. Bundled Python (for packaged apps)
3. System Python (Homebrew, etc.)

## Testing
Tested on macOS with Miniconda Python 3.13.2 configured in settings. The venv is now correctly created using the configured Python version instead of the system Python 3.9.6.

Co-Authored-By: Warp <agent@warp.dev>

## Summary by CodeRabbit

* **New Features**
  * Added support for a user-configurable Python path during startup and when saving settings.
  * The app validates and prioritizes a configured Python interpreter when preparing virtual environments.
  * If the configured path is invalid, the system falls back to existing Python detection and environment creation.


AndyMik90 (changes requested):
## 🤖 Auto Claude PR Review

7 issue(s) must be addressed (3 required, 4 recommended), 1 suggestions

### Risk Assessment
| Factor | Level | Notes |
|--------|-------|-------|
| Complexity | Low | Based on lines changed |
| Security Impact | Medium | Based on security findings |

### Findings Summary
- **High**: 3 issue(s)
- **Medium**: 4 issue(s)
- **Low**: 1 issue(s)

---

#### 🟠 [HIGH] Command injection risk: execSync with shell interpolation on user-configured path
📁 `apps/frontend/src/main/python-env-manager.ts:210`

The new findSystemPython() code uses execSync() with shell string interpolation: `execSync(`"${this.configuredPythonPath}" -c "import sys; print(sys.executable)"``. While validatePythonPath() filters many dangerous characters, execSync invokes a shell which can interpret sequences differently across platforms. The validation uses execFileSync (shell: false) but execution uses execSync (shell: true) - this inconsistency is a security anti-pattern.

**Suggested fix:**
```
Use execFileSync with shell: false instead: `execFileSync(this.configuredPythonPath, ['-c', 'import sys; print(sys.executable)'], { stdio: 'pipe', timeout: 5000, shell: false })`. Or use parsePythonCommand() helper as done elsewhere in the codebase.
```

#### 🟠 [HIGH] Missing Python version validation (3.10+ requirement not enforced)
📁 `apps/frontend/src/main/python-env-manager.ts:45`

The configure() method uses validatePythonPath() which only verifies the path is a valid Python executable, but does NOT check version meets the 3.10+ requirement. Users could configure Python 3.8 or 3.9, which would be accepted, but claude-agent-sdk requires 3.10+ and would fail at runtime. The existing findPythonCommand() uses validatePythonVersion() which enforces 3.10+, but this is bypassed for configured paths.

**Suggested fix:**
```
After validatePythonPath() succeeds in configure(), also call validatePythonVersion() from python-detector.ts to verify the version meets 3.10+ requirement. Reject with clear error if version is too old.
```

#### 🟠 [HIGH] Race condition: configure() may not complete before initialize() starts
📁 `apps/frontend/src/main/index.ts:205`

The PR adds pythonEnvManager.configure() in index.ts after agentManager.configure(), but pythonEnvManager.initialize() is called asynchronously from project-handlers.ts. There's no guarantee configure() completes before initialize() starts. If initialize() calls findSystemPython() before configure() sets configuredPythonPath, the user's configured path will be ignored for the first venv creation.

**Suggested fix:**
```
Ensure pythonEnvManager.configure() is called before any initialize() invocation. Consider passing pythonPath directly to initialize(), or call configure() earlier in the startup sequence before async handlers are registered.
```

#### 🟡 [MEDIUM] TOCTOU: Path validated once but used later
📁 `apps/frontend/src/main/python-env-manager.ts:45`

The configure() method validates the Python path once at configuration time, but stores it for later use in findSystemPython(). Between validation and execution, the file could be replaced with a malicious executable, especially if the path points to a writable location. This is a Time-of-Check to Time-of-Use vulnerability.

**Suggested fix:**
```
Re-validate the path immediately before use in findSystemPython(), or restrict allowed paths to system-protected locations that require elevated privileges to modify.
```

#### 🟡 [MEDIUM] State not invalidated when configure() called after initialize()
📁 `apps/frontend/src/main/python-env-manager.ts:45`

If initialize() has completed (isReady=true, pythonPath set), calling configure() afterwards has no effect on the already-created venv. The class state is not updated or invalidated. Users changing their Python path setting after initial setup will see no effect until app restart.

**Suggested fix:**
```
Add logic to configure() to either: warn if already initialized and path differs, invalidate current state requiring re-initialization, or document that configure() must be called before initialize().
```

#### 🟡 [MEDIUM] Silent fallback undermines user configuration intent
📁 `apps/frontend/src/main/python-env-manager.ts:215`

When configured Python path fails the execSync check in findSystemPython(), it logs an error and silently falls back to auto-detection. The user configured a specific Python for a reason (e.g., to avoid system Python 3.9.6). Silently using a different Python defeats the purpose and could lead to subtle version-related bugs.

**Suggested fix:**
```
When configured path fails: emit an error event so UI can notify the user, or throw an error instead of silent fallback, or at minimum clear configuredPythonPath so it's not retried on subsequent calls.
```

#### 🟡 [MEDIUM] Inconsistent verification: --version vs sys.executable
📁 `apps/frontend/src/main/python-env-manager.ts:50`

The configured path is verified twice with different methods: configure() uses validatePythonPath() which runs '--version', while findSystemPython() uses 'import sys; print(sys.executable)'. These test different things. A Python installation could pass --version but fail sys module import, creating confusing failure scenarios.

**Suggested fix:**
```
Use the same validation in both places, or remove the redundant check in findSystemPython() since configure() already validated. If keeping both, use consistent verification methods.
```

#### 🔵 [LOW] CodeRabbit: configure() should return validation status
📁 `apps/frontend/src/main/python-env-manager.ts:50`

The configure() method validates the Python path but returns void, providing no feedback to callers about whether validation succeeded. While the error is logged, callers have no programmatic way to know if configuration was accepted.

**Suggested fix:**
```
Change return type to boolean or a status object to indicate whether the configured path was accepted.
```

---

AndyMik90:

PR #526, opened 2026-01-01
fix: resolve frontend lag and update dependencies

AndyMik90:
## Summary

- Fix 2-5 second UI lag when multiple tasks are running simultaneously
- Batch IPC log events to reduce state updates from 100+/sec to ~6/sec
- Update all frontend dependencies including major versions

### Performance Fixes (commit 980323e7)

**Frontend optimizations:**
- Batch IPC log events using 16ms window with `unstable_batchedUpdates`
- Add `batchAppendLogs` to task-store for efficient log appending
- Only set `updatedAt` on phase changes, not every progress tick
- Memoize `sanitizeMarkdownForDisplay` and `formatRelativeTime` in TaskCard
- Add React.memo with custom comparators to DroppableColumn
- Use IntersectionObserver to pause animations when cards not visible

**Backend optimizations:**
- Add project index caching with 5-minute TTL in client.py
- Batch StatusManager file writes with threading.Timer debounce

### Dependency Updates (commit 083e6599)

| Package | Before | After |
|---------|--------|-------|
| react-resizable-panels | 3.0.6 | 4.2.0 |
| globals | 16.5.0 | 17.0.0 |
| lucide-react | 0.560.0 | 0.562.0 |
| zod | 4.2.1 | 4.3.4 |

Updated TerminalGrid.tsx for react-resizable-panels v4 API changes.

## Test plan

- [ ] Run app with multiple tasks (5+) simultaneously
- [ ] Verify UI remains responsive during task execution
- [ ] Verify terminal panel resizing works correctly
- [ ] Verify animations are smooth without choppy behavior

## Summary by CodeRabbit

* **New Features**
  * In-process project-index caching for faster client creation.
  * IPC batching and a public batched log-append API to reduce update churn.
  * Debounced, batched status writes with flush control.
  * Option to preserve new-commits check when storing PR review results.

* **Performance Improvements**
  * Widespread memoization to reduce re-renders.
  * Non-blocking atomic file persistence for responsiveness.

* **Bug Fixes**
  * Safer, race-resistant task persistence and more robust recovery/error handling.

* **Dependencies**
  * Updated lucide-react, react-resizable-panels, and globals.

### DIFF
--- a/apps/frontend/src/main/ipc-handlers/github/pr-handlers.ts
+++ b/apps/frontend/src/main/ipc-handlers/github/pr-handlers.ts
@@ -1341,25 +1341,32 @@ export function registerPRHandlers(
           return { hasNewCommits: false, newCommitCount: 0 };
         }
 
+        // Fetch PR data to get current HEAD (before try block so it's accessible in catch)
+        let currentHeadSha: string;
         try {
-          // Get PR data to find current HEAD
           const prData = (await githubFetch(
             config.token,
             `/repos/${config.repo}/pulls/${prNumber}`
           )) as { head: { sha: string }; commits: number };
+          currentHeadSha = prData.head.sha;
+        } catch (error) {
+          debugLog('Error fetching PR data', { prNumber, error: error instanceof Error ? error.message : error });
+          return { hasNewCommits: false, newCommitCount: 0 };
+        }
 
-          const currentHeadSha = prData.head.sha;
+        // Early return if SHAs match - no new commits
+        if (reviewedCommitSha === currentHeadSha) {
+          return {
+            hasNewCommits: false,
+            newCommitCount: 0,
+            lastReviewedCommit: reviewedCommitSha,
+            currentHeadCommit: currentHeadSha,
+            hasCommitsAfterPosting: false,
+          };
+        }
 
-          if (reviewedCommitSha === currentHeadSha) {
-            return {
-              hasNewCommits: false,
-              newCommitCount: 0,
-              lastReviewedCommit: reviewedCommitSha,
-              currentHeadCommit: currentHeadSha,
-              hasCommitsAfterPosting: false,
-            };
-          }
-
+        // Try to get detailed comparison
+        try {
           // Get comparison to count new commits
           const comparison = (await githubFetch(
             config.token,
@@ -1396,8 +1403,21 @@ export function registerPRHandlers(
             hasCommitsAfterPosting,
           };
         } catch (error) {
-          debugLog('Error checking new commits', { prNumber, error: error instanceof Error ? error.message : error });
-          return { hasNewCommits: false, newCommitCount: 0 };
+          // Comparison failed (e.g., force push made old commit unreachable)
+          // Since we already verified SHAs differ, treat as having new commits
+          debugLog('Comparison failed but SHAs differ - likely force push, treating as new commits', {
+            prNumber,
+            reviewedCommitSha,
+            currentHeadSha,
+            error: error instanceof Error ? error.message : error
+          });
+          return {
+            hasNewCommits: true,
+            newCommitCount: 1, // Unknown count due to force push
+            lastReviewedCommit: reviewedCommitSha,
+            currentHeadCommit: currentHeadSha,
+            hasCommitsAfterPosting: true, // Assume yes for force push scenarios
+          };
         }
       });
 
--- a/apps/frontend/src/renderer/components/github-prs/components/PRDetail.tsx
+++ b/apps/frontend/src/renderer/components/github-prs/components/PRDetail.tsx
@@ -99,11 +99,10 @@ export function PRDetail({
   const [isLoadingLogs, setIsLoadingLogs] = useState(false);
   const logsLoadedRef = useRef(false);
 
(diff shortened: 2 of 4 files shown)


AndyMik90 (commented):
## 🤖 Auto Claude PR Review

6 issue(s) must be addressed (1 required, 5 recommended), 2 suggestions

### Risk Assessment
| Factor | Level | Notes |
|--------|-------|-------|

### Findings Summary
- **High**: 1 issue(s)
- **Medium**: 5 issue(s)
- **Low**: 2 issue(s)

---

#### 🟠 [HIGH] Missing error handling for writeFileSync operations
📁 `apps/frontend/src/main/ipc-handlers/task/execution-handlers.ts:304`

Multiple writeFileSync calls (lines 304-308, 364-366, 690, 735-786) lack try-catch blocks. If the filesystem is unavailable, disk is full, or permissions are wrong, these will throw unhandled exceptions and crash the handler. QA report writing and recovery plan updates are affected.

**Suggested fix:**
```
Wrap file write operations in try-catch: `try { writeFileSync(path, content); } catch (error) { console.error('Failed to write file:', error); return { success: false, error: 'Failed to write file' }; }`
```

#### 🟡 [MEDIUM] Race condition in debounced write timer management
📁 `apps/backend/ui/status.py:115`

The _write_pending and _write_timer instance variables are accessed from multiple threads without synchronization. The Timer thread executes _do_write() which sets _write_pending=False (line 139) while _schedule_write() can be called simultaneously from the main thread. This can cause lost updates if timing aligns incorrectly - _do_write clears the pending flag just as _schedule_write sets it.

**Suggested fix:**
```
Add a threading.Lock to protect _write_pending and _write_timer access: `self._lock = threading.Lock()` in __init__, then wrap all accesses in `with self._lock:` blocks.
```

#### 🟡 [MEDIUM] Race condition in global cache without thread synchronization
📁 `apps/backend/core/client.py:30`

The global _PROJECT_INDEX_CACHE dict is accessed and modified without locking. The check-then-act pattern in _get_cached_project_data (lines 52-80) is not atomic: Thread A can check if key exists, Thread B can delete that key via invalidate_project_cache(), causing Thread A to get a KeyError. Two threads can also both have a cache miss and both load the project index, wasting work.

**Suggested fix:**
```
Add a threading.Lock: `_CACHE_LOCK = threading.Lock()` and use `with _CACHE_LOCK:` around the check-read-write operations in _get_cached_project_data() and invalidate_project_cache().
```

#### 🟡 [MEDIUM] Batch timeout not cancelled on component unmount
📁 `apps/frontend/src/renderer/hooks/useIpc.ts:89`

When the component unmounts, the batchTimeout is not cleared. If there are queued IPC updates and the 16ms hasn't elapsed, flushBatch() will execute after unmount. Additionally, the module-level storeActions variable gets overwritten on each hook render, which could cause stale reference bugs if multiple components use the hook.

**Suggested fix:**
```
In the useEffect cleanup, add: `if (batchTimeout) { clearTimeout(batchTimeout); flushBatch(); batchTimeout = null; }` to flush pending updates and clear the timeout on unmount. Consider using useRef for storeActions instead of module-level state.
```

#### 🟡 [MEDIUM] IntersectionObserver recreated on every visibility change
📁 `apps/frontend/src/renderer/components/PhaseProgressIndicator.tsx:60`

The useEffect dependency array includes isVisible (line 80), causing the IntersectionObserver to be disconnected and reconnected on every visibility change. The observer callback sets isVisible, which triggers a re-render and recreates the observer, creating an unnecessary cycle. This wastes resources and causes the wasVisible comparison to be misleading.

**Suggested fix:**
```
Remove isVisible from the dependency array and use a ref to track previous visibility: `const prevVisibleRef = useRef(true); useEffect(() => { prevVisibleRef.current = isVisible; }, [isVisible]);` Then use prevVisibleRef.current instead of wasVisible in the observer callback.
```

#### 🟡 [MEDIUM] Inline arrow function in memoized context creates new reference every render
📁 `apps/frontend/src/renderer/components/KanbanBoard.tsx:153`

Inside the memoized taskCards (lines 147-155), each SortableTaskCard receives `onClick={() => onTaskClick(task)}` which creates a new function reference. This defeats the purpose of the custom memo comparator in SortableTaskCard which checks onClick reference equality. Every taskCard will re-render on any parent update because onClick is always a new reference.

**Suggested fix:**
```
Create stable callbacks: `const onClickHandlers = useMemo(() => new Map(tasks.map(t => [t.id, () => onTaskClick(t)])), [tasks, onTaskClick]);` then use `onClick={onClickHandlers.get(task.id)}`. Alternatively, pass onTaskClick and task separately.
```

#### 🔵 [LOW] Unused import: useCallback
📁 `apps/frontend/src/renderer/components/KanbanBoard.tsx:1`

useCallback is imported from React but never used in the component. The inline arrow functions should either use useCallback or the import should be removed.

**Suggested fix:**
```
Remove useCallback from the import: `import { useState, useMemo, useEffect } from 'react';`
```

#### 🔵 [LOW] Unused imports: useCallback, useRef
📁 `apps/frontend/src/renderer/hooks/useIpc.ts:1`

useCallback and useRef are imported from React but never used in the hook. These should be removed for cleaner code.

**Suggested fix:**
```
Remove unused imports: `import { useEffect } from 'react';`
```

---

PR #549, opened 2026-01-02
fix(csp): allow external HTTPS images in Content-Security-Policy

MikeeBuilds:
Updates img-src CSP directive to include https: allowing external images (like Supabase Storage) to load in GitHub issue previews.

## Summary by CodeRabbit

* **Chores**
  * Updated Content-Security-Policy to allow loading images from secure external HTTPS sources (in addition to data and blob), enabling support for securely hosted external images.


adryserage:
## Security Enhancement Suggestion

Thanks for this fix! The reviewers are right that `https:` is too permissive. Based on a codebase search, the app only needs images from specific sources:

### Recommended CSP change:

```html
img-src 'self' data: blob: https://avatars.githubusercontent.com https://*.githubusercontent.com
```

This allows:
- GitHub user/org avatars (`avatars.githubusercontent.com`)
- GitHub-hosted images (`*.githubusercontent.com` for raw content, user content, etc.)

If Supabase Storage is also needed, add:
```
https://*.supabase.co
```

This is much more secure than allowing all HTTPS sources while still supporting the use cases mentioned in the PR description.

PR #748, opened 2026-01-06
fix(frontend): ensure PATH includes system directories when launched

czerewacz:
Fixes [ACS-105](https://linear.app/mikee-builds-tech/issue/ACS-105/bug-insights-feature-fails-with-process-exited-with-code-1-missing) (Linear issue)

  ## Investigation Summary

  ### Original Bug (ACS-105): ✅ ALREADY FIXED

  **Issue:** "Process exited with code 1" - Python not finding modules

  **Root cause:** Missing PYTHONPATH in packaged DMG

  **Fix:** Commit f5be794 (January 3, 2026) by @hluisi - added `pythonEnv` and `getAugmentedEnv()` to Insights environment

  **Status:** Fixed in `develop`, will be in v2.7.3 release

  **Verified:** Built and tested DMG from fork - original crash no longer occurs ✅

  ### New Bug Discovered During Testing: ⚠️ SDK FAILURE → TIMEOUT

  **Issue:** Even simple "hello" query times out after 120s in packaged DMG

  **CRITICAL DISCOVERY:**
  - ✅ Works when launched from terminal
  - ❌ Times out when double-clicked from Finder

  **Root Cause Analysis:**
  - Terminal launch: Inherits shell environment (PATH includes all system directories from `.zshrc`)
  - Finder/Dock launch: macOS provides minimal environment, PATH often empty or missing `/usr/bin`
  - Impact: Claude Agent SDK cannot find `/usr/bin/security` to access macOS Keychain for OAuth tokens
  - Result: SDK initialization fails, falls back to simple mode with 120s timeout

  **Affected users:** Anyone who launches Auto-Claude normally (double-click) instead of from terminal

  ## Solution

  Ensure PATH always includes essential system directories in `getAugmentedEnv()`.

  ## Changes

  - **apps/frontend/src/main/env-utils.ts**:
    - Add check for essential paths: `/usr/bin`, `/bin`, `/usr/sbin`, `/sbin`
    - Only appends missing paths (respects user's existing PATH)
    - Applies to macOS and Linux (not Windows)

  ## Testing

  1. Built DMG from this branch: `npm run package:mac`
  2. Launched by **double-clicking** (not from terminal)
  3. Opened Insights tab and sent "hello" query
  4. ✅ **Result:** Received response without timeout
     - Previously: Would timeout after 120s
     - Now: Responds immediately using Claude Agent SDK

  ## Base Branch

  - [x] This PR targets the `develop` branch (required for all feature/fix PRs)

  ## Type of Change

  - [x] 🐛 Bug fix

  ## Area

  - [x] Frontend

  ## Checklist

  - [x] I've synced with `develop` branch
  - [x] I've tested my changes locally
  - [x] I've followed the code principles (SOLID, DRY, KISS)
  - [x] My PR is small and focused (< 400 lines ideally)

  ## Feature Toggle

  - [x] N/A - Feature is complete and ready for all users

  ## Breaking Changes

  **Breaking:** No

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures essential system directories are present in PATH on non-Windows systems and consistently updates PATH using the augmented value so runtime tools are reliably discoverable.

* **Documentation**
  * Clarified comments describing essential system paths and platform behavior for maintainers.

### DIFF
--- a/apps/frontend/src/main/env-utils.ts
+++ b/apps/frontend/src/main/env-utils.ts
@@ -360,9 +360,22 @@ export async function getAugmentedEnvAsync(additionalPaths?: string[]): Promise<
   // Get all candidate paths (platform + additional)
   const candidatePaths = getExpandedPlatformPaths(additionalPaths);
 
+  // Ensure essential system paths are present (for macOS Keychain access)
+  let currentPath = env.PATH || '';
+
+  if (platform !== 'win32') {
+    const pathSetForEssentials = new Set(currentPath.split(pathSeparator).filter(Boolean));
+    const missingEssentials = ESSENTIAL_SYSTEM_PATHS.filter(p => !pathSetForEssentials.has(p));
+
+    if (missingEssentials.length > 0) {
+      currentPath = currentPath
+        ? `${currentPath}${pathSeparator}${missingEssentials.join(pathSeparator)}`
+        : missingEssentials.join(pathSeparator);
+    }
+  }
+
   // Collect paths to add (only if they exist and aren't already in PATH)
-  const currentPath = env.PATH || '';
-  const currentPathSet = new Set(currentPath.split(pathSeparator));
+  const currentPathSet = new Set(currentPath.split(pathSeparator).filter(Boolean));
 
   // Check existence asynchronously in parallel for performance
   const pathChecks = await Promise.all(
@@ -382,9 +395,7 @@ export async function getAugmentedEnvAsync(additionalPaths?: string[]): Promise<
   const pathsToAdd = buildPathsToAdd(candidatePaths, currentPathSet, existingPaths, npmPrefix);
 
   // Prepend new paths to PATH (prepend so they take priority)
-  if (pathsToAdd.length > 0) {
-    env.PATH = [...pathsToAdd, currentPath].filter(Boolean).join(pathSeparator);
-  }
+  env.PATH = [...pathsToAdd, currentPath].filter(Boolean).join(pathSeparator);
 
   return env;
 }


apps/frontend/src/main/env-utils.ts
```diff
@@ -109,8 +109,26 @@ export function getAugmentedEnv(additionalPaths?: string[]): Record<string, stri
     p.startsWith('~') ? p.replace('~', homeDir) : p
   );
 
+  // Ensure PATH has essential system directories when launched from Finder/Dock.
+  // When Electron launches from GUI (not terminal), PATH might be empty or minimal.
+  // The Claude Agent SDK needs /usr/bin/security to access macOS Keychain.
+  let currentPath = env.PATH || '';
+
+  // On macOS/Linux, ensure basic system paths are always present
+  if (platform !== 'win32') {
```
dertuerke, line 210: Add currentpath also to the if condition. Empty string are falsely in JavaScript.

czerewacz: @dertuerke Thanks for the review! I want to make sure I understand your concern correctly.

 The current code handles empty PATH by:
 1. Line 115: `let currentPath = env.PATH || '';` - defaults to empty string
 2. Line 120: `.filter(Boolean)` - removes empty strings after split
 3. Line 125-127: Ternary handles both empty and non-empty cases

 When PATH is empty (which is the bug being fixed), the code correctly sets it to `/usr/bin:/bin:/usr/sbin:/sbin`.

 Could you please clarify what specific issue you're seeing? Are you suggesting:
 - A performance optimization (skip Set operations when PATH is empty)?
 - A code clarity improvement (make empty PATH handling more explicit)?
 - Something else I'm missing?

 Happy to improve the code - just want to make sure I address your concern correctly!

czerewacz: @dertuerke You were absolutely right to be concerned about empty string handling!

  The AI review found the actual issue: inconsistent `.filter(Boolean)` usage. I've updated the code to:

  1. ✅ Add `.filter(Boolean)` to line 138 for consistency (both `currentPathSet` creations now filter empty strings)
  2. ✅ Extract the essential paths array to a module-level constant (`ESSENTIAL_SYSTEM_PATHS`)

  This properly handles empty strings throughout the function and follows the file's existing patterns. Thanks for catching this! 🙏

dertuerke: Sorry, I described it a bit too vaguely. What I meant to say is, if the path is empty or undefined, you're setting an empty string. Empty strings are always false in JavaScript.

czerewacz: Thanks @dertuerke! 
I understand that when env.PATH is undefined/null, we set currentPath = '' (empty string), and empty strings are falsy in JavaScript.
In this case, the code is designed to handle this case correctly.

  Let me trace through what happens when PATH is empty:

  let currentPath = env.PATH || '';  _// currentPath = '' (empty string)_

  if (platform !== 'win32') {  
    // Split empty string: ''.split(':') = ['']
    // Filter removes it: [''].filter(Boolean) = []
    const pathSetForEssentials = new Set([]);  // Empty set

    // All essentials are "missing" since Set is empty
    const missingEssentials = ESSENTIAL_SYSTEM_PATHS.filter(...);  // All 4 paths

    if (missingEssentials.length > 0) {  // TRUE, we have 4 missing
      currentPath = currentPath  // empty string is falsy, so...
        ? `...`  // ← SKIP this branch
        : missingEssentials.join(':');  // ← Use THIS branch ✓
      // Result: currentPath = '/usr/bin:/bin:/usr/sbin:/sbin'
    }
  }

  The empty string being falsy is CORRECT behavior! It makes the ternary operator choose the right branch.
  
  Then, on windows, the essential paths block is intentionally skipped because Windows doesn't have the same problem as macOS
  
  On Windows:
  1. The if (platform !== 'win32') check fails, so the entire essential paths block is skipped
  2. currentPath stays as env.PATH || '' (empty string is fine)
  3. Only the common tool paths from COMMON_BIN_PATHS.win32 are added if they exist:
    - C:\Program Files\Git\cmd
    - C:\Program Files\GitHub CLI
  
  
 Let me know please if Im missing something :D

apps/frontend/src/main/env-utils.ts
```diff
@@ -212,9 +235,7 @@ export function getAugmentedEnv(additionalPaths?: string[]): Record<string, stri
   const pathsToAdd = buildPathsToAdd(candidatePaths, currentPathSet, existingPaths, npmPrefix);
 
   // Prepend new paths to PATH (prepend so they take priority)
-  if (pathsToAdd.length > 0) {
-    env.PATH = [...pathsToAdd, currentPath].filter(Boolean).join(pathSeparator);
-  }
+  env.PATH = [...pathsToAdd, currentPath].filter(Boolean).join(pathSeparator);
```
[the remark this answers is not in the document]
czerewacz: Commited fix based on this comment:
1. Added essential paths logic to getAugmentedEnvAsync() (lines 363-375)
    - Same as sync version: ensures /usr/bin, /bin, etc. are in PATH
    - Critical for macOS Keychain access in async code paths
  2. Added .filter(Boolean) to currentPathSet (line 378)
    - Prevents empty string from being treated as a path entry
    - Ensures correct duplicate detection
  3. Removed conditional PATH update (line 398)
    - Changed from if (pathsToAdd.length > 0) to unconditional
    - Ensures essential paths are written even on minimal systems

AndyMik90 (approved):

PR #753, opened 2026-01-06
fix(windows): kill process tree on app close to prevent stale processes

Orinks:
## Summary

- Fixes stale process issue on Windows where PTY daemon and child processes weren't properly cleaned up when Auto Claude UI closes
- Windows doesn't propagate signals to child processes like Unix does, so `process.kill()` only killed the parent, leaving orphan `python.exe`, `cmd.exe`, `pwsh.exe` processes

## Changes

- Add `killProcessTree()` helper in `agent-process.ts` that uses `taskkill /PID {pid} /T /F` on Windows
- Update `killProcess()` to use the helper for Windows process tree termination
- Update `killPty()` in `pty-manager.ts` to kill the PTY process tree on Windows before calling `pty.kill()`

## Test plan

- [ ] Start Auto Claude on Windows, open a terminal, run a long command
- [ ] Close the app
- [ ] Open Task Manager → verify no orphan `cmd.exe`, `pwsh.exe`, `python.exe` processes
- [ ] Repeat with a build task running

## Summary by CodeRabbit

* **New Features**
  * Added a cross-platform process termination utility for reliable stopping of processes and descendants.

* **Bug Fixes**
  * Platform-aware termination: immediate on Windows; graceful then forced on Unix with a fallback timer.
  * Ensure child processes are killed before terminal/PTy cleanup.
  * Immediate state cleanup to avoid stale process references and guards against invalid PIDs.
  * Force-kill timer is canceled when processes exit to avoid redundant kills.

### DIFF
--- a/apps/frontend/src/main/utils/process-utils.ts
+++ b/apps/frontend/src/main/utils/process-utils.ts
@@ -8,9 +8,13 @@ import { spawnSync } from 'child_process';
 const TASKKILL_TIMEOUT_MS = 5000;
 
 /**
- * Kill a process and all its children (process tree).
- * On Windows, uses taskkill /T to kill the entire process tree.
- * On Unix, uses standard signal-based killing.
+ * Kill a process tree on Windows, or signal a process on Unix.
+ *
+ * On Windows, uses taskkill /T to kill the entire process tree (parent + all children).
+ * On Unix, sends a signal to the specified process only. This does NOT kill the full
+ * process tree - child processes may continue running (though well-behaved children
+ * typically exit when their parent dies). Full Unix process tree killing would require
+ * spawning with detached:true and using negative PIDs, which is out of scope here.
  *
  * @param pid - Process ID to kill
  * @param signal - Signal to send (Unix only, Windows always force-kills)


Orinks:
## Re: Unix process tree killing

Thanks for the review feedback! The issues raised by CodeRabbit have been addressed in commit 75dda71:
- ✅ Added 5s timeout to `spawnSync` to prevent indefinite blocking
- ✅ Fixed stale reference by capturing `child` process before deleting from state
- ✅ Added `exit` listener to clear force-kill timer on graceful exit
- ✅ Extracted to shared `process-utils.ts` utility to remove duplication

Regarding the Unix process tree killing concern from @gemini-code-assist:

The current implementation intentionally only signals the parent process on Unix, not the full process tree. This is because:

1. **Scope**: This PR specifically targets the Windows stale process issue where `process.kill()` doesn't terminate child processes
2. **Unix behavior**: On Unix, well-behaved child processes typically exit when their parent dies (or receive SIGHUP). The current signal-based approach works for our use case
3. **Breaking changes**: Using `process.kill(-pid, signal)` would require spawning processes with `detached: true`, which could have unintended side effects on agent behavior

I've updated the docstring in `process-utils.ts` to clarify this limitation. If full Unix process tree killing is needed in the future, it can be addressed in a separate PR with proper testing of the `detached` spawn option.

Orinks:
I think this has already been fixed on develop branch.

PR #953, opened 2026-01-12
fix(ci): fix semver comparison for release workflow

AndyMik90:
## Summary

- Fix `prepare-release.yml` to use `npx semver` for proper semantic version comparison
  - `sort -V` incorrectly ranked `2.7.3-beta.1` > `2.7.3`, blocking the release
  - Now correctly recognizes that stable versions are newer than pre-release versions
- Add `workflow_dispatch` trigger with force option for manual releases
- Fix CHANGELOG.md formatting (remove `#` that caused header rendering issue)

## Why This Was Needed

The v2.7.3 release was blocked because the version comparison logic used `sort -V` which doesn't follow semver rules for pre-release versions. In semver, `2.7.3` (stable) should be greater than `2.7.3-beta.1` (pre-release), but `sort -V` sorted them alphabetically.

## Test Plan

- [ ] Merge this PR to main
- [ ] Manually trigger "Prepare Release" workflow from Actions tab
- [ ] Verify v2.7.3 tag is created and release workflow starts


AndyMik90:
Closing - need to sync develop with main first after squash merge

Issue #981, opened 2026-01-12
Task Moves to "Human Review" While Validation Loop Is Still Running


coygeek:
### Checklist

- [x] I searched existing issues and this hasn't been reported

### Area

Frontend

### Operating System

macOS

### Version

2.7.3 official release

### What happened?

When clicking "Refresh Tasks" in the kanban board, a task prematurely moves to the "Human Review" column while the QA validation loop is still actively running (visible in the logs tab as ongoing validation entries).

The task progression was:
1. Planning phase completed (53 entries)
2. Coding phase completed (117 entries)
3. Moved to "AI Review" (QA validation started)
4. **Bug:** Clicked "Refresh Tasks" → task moved to "Human Review"
5. But the Logs tab showed validation was still running (49 entries in progress)

## Root Cause Analysis

The bug is in the status determination logic in `ImplementationPlan.update_status_from_subtasks()`:

**File:** `apps/backend/models/plan.py` (around line 33711)

```python
def update_status_from_subtasks(self):
    # ...
    if completed_count == total_count:
        # All subtasks completed - check if QA approved
        if self.qa_signoff and self.qa_signoff.get("status") == "approved":
            self.status = "human_review"
            self.planStatus = "review"
        else:
            # All subtasks done, waiting for QA
            self.status = "ai_review"
            self.planStatus = "review"
```

**The Problem:**
The code checks two conditions to move to "human_review":
1. All subtasks completed (`completed_count == total_count`)
2. `qa_signoff.status == "approved"`

However, it does **NOT** check whether the QA validation loop is still actively running.

**What Happens:**
1. All coding subtasks complete → `completed_count == total_count` is true
2. QA validation loop starts (can run 49+ iterations)
3. During an iteration, the QA agent updates `implementation_plan.json` with `qa_signoff.status = "approved"`
4. The validation loop may still be running (checking additional criteria, running fixes, etc.)
5. User clicks "Refresh Tasks" → frontend calls backend API
6. Backend reads `implementation_plan.json` and calls `update_status_from_subtasks()`
7. Method sees: subtasks complete + qa_signoff approved → sets `status = "human_review"`
8. Kanban shows task in "Human Review" column
9. **BUT the validation loop is still actively running in the background**

**Related Code Locations:**
- `apps/backend/qa_loop/criteria.py` - `is_qa_approved()` function (line ~73261)
- `apps/backend/qa_loop/__init__.py` - `run_qa_validation_loop()` function (line ~73778)
[...]

PR #1017, opened 2026-01-13
Fix Task Premature Status Transition During QA Validation Loop

AndyMik90:
This bug fix addresses GitHub Issue #981 where tasks prematurely move to "Human Review" status when users click "Refresh Tasks" while the QA validation loop is still actively running in the background. The root cause is a race condition in the status determination logic (`apps/backend/models/plan.py`) that checks only subtask completion and QA approval status, but fails to verify whether the validation loop has fully terminated. The fix implements a `validation_complete` flag to ensure tasks only transition after the QA loop completes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed task status transitions that were occurring prematurely during QA validation. Tasks now correctly remain in review status until QA validation is fully complete before advancing to the next stage, ensuring a more reliable workflow.


AndyMik90:
Closing PR - branch history corrupted with unrelated changes. Will restart with clean implementation.

PR #1185, opened 2026-01-16
Fix Mac Crash on Invoke Claude Button

AndyMik90:
The Auto Claude desktop application crashes on macOS when users click the "Invoke Claude" button in agent terminals. The root cause is synchronous file I/O operations (`writeFileSync`) in the terminal session persistence layer that block the Electron main process. This task converts these blocking operations to asynchronous, fire-and-forget persistence to prevent UI thread blocking and eliminate the crash.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,7 +2,7 @@
 
 # Preserve git worktree context - prevent HEAD corruption in worktrees
 if [ -f ".git" ]; then
-  WORKTREE_GIT_DIR=$(cat .git | sed 's/gitdir: //')
+  WORKTREE_GIT_DIR=$(sed 's/^gitdir: //' .git)
   if [ -n "$WORKTREE_GIT_DIR" ]; then
     export GIT_DIR="$WORKTREE_GIT_DIR"
     export GIT_WORK_TREE="$(pwd)"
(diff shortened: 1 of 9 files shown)


AndyMik90 (commented):
## 🤖 Auto Claude PR Review

## 🟠 Follow-up Review: Needs Revision

> **🟠 Needs revision** - 1 blocking issue(s) require fixes.

### Resolution Status
- ✅ **Resolved**: 6 previous findings addressed
- ❌ **Unresolved**: 0 previous findings remain
- 🆕 **New Issues**: 1 new findings in recent changes

### Finding Validation
- 🔍 **Dismissed as False Positives**: 4 findings were re-investigated and found to be incorrect
- ✓ **Confirmed Valid**: 1 findings verified as genuine issues
- 👤 **Needs Human Review**: 1 findings require manual verification

### 🚨 Blocking Issues
- quality: persistAllSessionsAsync may not await final disk write during shutdown

### Verdict
CI Status: 2 checks still pending - cannot approve until CI completes. Previous Findings: All 6 previous findings have been successfully addressed (5 resolved, 1 dismissed as false positive after validation). The code changes demonstrate excellent work addressing race conditions, code duplication, and async persistence. New Issues: 1 confirmed MEDIUM severity finding (NEW-001) about potential data loss during app shutdown due to write coalescing - when Promise.all() in persistAllSessionsAsync resolves, the final disk write may still be pending via setImmediate. This edge case should be addressed for production safety, particularly since destroyAllTerminals is a critical shutdown path. 3 other new-code-reviewer findings were dismissed as out-of-scope (pre-existing issues in unchanged code).

### Review Process

---


### Findings (1 selected of 1 total)

#### 🟡 [NEW-001] [MEDIUM] persistAllSessionsAsync may not await final disk write during shutdown
📁 `apps/frontend/src/main/terminal/session-handler.ts:200`

The Promise.all() in persistAllSessionsAsync may resolve before the final coalesced disk write completes. When multiple saveSessionAsync calls happen in parallel, write coalescing causes intermediate calls to return immediately (setting writePending=true). The final write is scheduled via setImmediate(), which executes AFTER Promise.all() resolves. During app shutdown in destroyAllTerminals, the process may exit before this pending write completes.

**Suggested fix:**
```
Consider adding a flushAsync() method that waits for any pending/in-progress writes to complete, and call it at the end of persistAllSessionsAsync for critical shutdown paths. Alternatively, track a writeComplete promise that resolves only when disk write finishes.
```

---

PR #1771, opened 2026-02-09
fix(fast-mode): use setting_sources instead of env var for CLI fast mode

AndyMik90:
## Summary

Fast Mode was non-functional because the `CLAUDE_CODE_FAST_MODE` env var isn't read by the Claude Code CLI. The CLI reads `fastMode` from user settings (`~/.claude/settings.json`) via the `/fast` toggle. This PR fixes the SDK integration to match how the CLI actually works, and makes several related improvements.

## Problem

When a user enabled Fast Mode, Auto Claude set `CLAUDE_CODE_FAST_MODE=true` as an environment variable for the CLI subprocess. However, the Claude Code CLI ignores this env var — it only reads `fastMode` from `~/.claude/settings.json` when the `user` setting source is enabled. Since the Agent SDK defaults to `--setting-sources ""` (blocking all filesystem settings), Fast Mode was silently doing nothing.

## Solution

- **Write `fastMode=true` to `~/.claude/settings.json`** before spawning the CLI, using atomic file writes (`tempfile` + `os.replace`) to prevent corruption from concurrent task processes
- **Enable `setting_sources=["user"]`** in the SDK options so the CLI reads the user settings file
- **Register an `atexit` handler** to restore `fastMode=false` when the task process exits, preventing fast mode from leaking into interactive sessions
- **Move fast mode toggle from global settings to per-task** — each task now independently controls whether it uses fast mode, stored in `task_metadata.json`

### Backend (Python)
- **`core/fast_mode.py`** (NEW) — Extracted shared fast mode helpers (atomic write, atexit cleanup)
- **`core/client.py`** — Uses `setting_sources=["user"]` instead of env var; imports from `core/fast_mode`
- **`core/simple_client.py`** — Same fix, imports from shared `core/fast_mode`
- **`phase_config.py`** — Added `sanitize_thinking_level()` for legacy value mapping (`ultrathink`→`high`, `none`→`low`), added `opus-4.5` model
- **All runners** — Removed `argparse choices` constraint on `--thinking-level`, added post-parse `sanitize_thinking_level()` call
- **`requirements.txt`** — Bumped `claude-agent-sdk>=0.1.33` for Opus 4.6 adaptive thinking support

### Frontend (TypeScript/React)
- **Per-task fast mode toggle** in TaskCreationWizard, TaskEditDialog, TaskFormFields (shown when any phase uses Opus)
- **Fast Mode badge** on TaskCard when enabled
- **Adaptive thinking badge** on AgentProfileSettings (replaces global fast mode toggle)
- **`opus-4.5` model option** added throughout type system and UI
- **Shared constants** — Exported `LEGACY_THINKING_MAP`, `sanitizeThinkingLevel()`, `PHASE_KEYS`, `VALID_THINKING_LEVELS`, `ADAPTIVE_THINKING_MODELS` from `shared/constants/models.ts`
- **Settings migration** — `_migratedUltrathinkToHigh` converts stored legacy thinking levels
- **Thinking level sanitization** at task create/update and spec creation paths
- **`useIpc.ts`** — Fixed project filtering to use `activeProjectId || selectedProjectId`
- **`credential-utils.ts`** — Added `stdio: 'pipe'` to prevent keychain error leaking to stderr

### Tests & Scripts
- **`tests/test_thinking_level_validation.py`** — Added `TestSanitizeThinkingLevel` class (5 new tests)
- **`scripts/diagnostic_fast_mode_invocations.py`** — Moved from `tests/` to prevent pytest collection (manual diagnostic tool, not automated tests)

## Testing

- [x] Frontend: biome ci, typecheck, 3037 tests pass, build succeeds
- [x] Backend: ruff check, ruff format, 2189 pytest tests pass
- [x] Pre-PR validation with 9 parallel review agents + 3 fix agents + self-review
- [ ] Manual: verify fast mode activates (extra usage billing) with Opus 4.6 task

## Summary by CodeRabbit

* **New Features**
  * Added Opus 4.5 model and per-task Fast Mode toggle (persisted on tasks)

* **UI Improvements**
  * Fast Mode toggle in task creation/edit flows and Fast badge on task cards
  * Adaptive Thinking indicator added to agent profile settings

* **Behavior & Validation**
  * Thinking-level normalization (legacy values auto-mapped) and one‑time migration
  * Fast Mode handling moved from global app settings to per-task flows with safe CLI support

* **Localization**
  * English & French strings updated for Fast Mode and Adaptive Thinking

* **Tests**
  * New tests validating thinking-level sanitization

### DIFF
--- a/apps/backend/agents/coder.py
+++ b/apps/backend/agents/coder.py
@@ -591,6 +591,9 @@ async def run_autonomous_agent(
 
         # Generate appropriate prompt
         fast_mode = get_fast_mode(spec_dir)
+        logger.info(
+            f"[Coder] [Fast Mode] {'ENABLED' if fast_mode else 'disabled'} for phase={current_phase}"
+        )
 
         if first_run:
             # Create client for planning phase
--- a/apps/backend/agents/planner.py
+++ b/apps/backend/agents/planner.py
@@ -104,6 +104,9 @@ async def run_followup_planner(
         spec_dir, "planning", planning_model
     )
     fast_mode = get_fast_mode(spec_dir)
+    logger.info(
+        f"[Planner] [Fast Mode] {'ENABLED' if fast_mode else 'disabled'} for follow-up planning"
+    )
     client = create_client(
         project_dir,
         spec_dir,
--- a/apps/backend/core/client.py
+++ b/apps/backend/core/client.py
@@ -21,7 +21,7 @@ import time
 from pathlib import Path
 from typing import Any
 
-from core.fast_mode import _ensure_fast_mode_in_user_settings
+from core.fast_mode import ensure_fast_mode_in_user_settings
 from core.platform import (
     is_windows,
     validate_cli_path,
@@ -525,7 +525,7 @@ def create_client(
     # so the CLI reads ~/.claude/settings.json where fastMode: true lives.
     # See: https://code.claude.com/docs/en/fast-mode
     if fast_mode:
-        _ensure_fast_mode_in_user_settings()
+        ensure_fast_mode_in_user_settings()
         logger.info("[Fast Mode] ACTIVE — will enable user setting source for fastMode")
         print(
             "[Fast Mode] ACTIVE — enabling user settings source for CLI to read fastMode"
--- a/apps/backend/core/client.py
+++ b/apps/backend/core/client.py
@@ -21,6 +21,7 @@ import time
 from pathlib import Path
 from typing import Any
 
+from core.fast_mode import _ensure_fast_mode_in_user_settings
 from core.platform import (
     is_windows,
     validate_cli_path,
@@ -483,9 +484,10 @@ def create_client(
                      "medium", "high"). When set, injected as CLAUDE_CODE_EFFORT_LEVEL
                      env var for the SDK subprocess. Only meaningful for models that
                      support adaptive thinking (e.g., Opus 4.6).
-        fast_mode: Enable Fast Mode for faster Opus 4.6 output. When True, injected
-                  as CLAUDE_CODE_FAST_MODE=true env var. Requires extra usage enabled
-                  on Claude subscription; falls back to standard speed automatically.
+        fast_mode: Enable Fast Mode for faster Opus 4.6 output. When True, enables
+                  the "user" setting source so the CLI reads fastMode from
+                  ~/.claude/settings.json. Requires extra usage enabled on Claude
+                  subscription; falls back to standard speed automatically.
 
     Returns:
         Configured ClaudeSDKClient
@@ -517,9 +519,19 @@ def create_client(
     if effort_level:
         sdk_env["CLAUDE_CODE_EFFORT_LEVEL"] = effort_level
 
-    # Inject fast mode for faster Opus 4.6 output
+    # Fast mode requires the CLI to read "fastMode" from user settings.
+    # The SDK default (setting_sources=None) passes --setting-sources "" which
+    # blocks ALL filesystem settings. We must explicitly enable "user" source
+    # so the CLI reads ~/.claude/settings.json where fastMode: true lives.
+    # See: https://code.claude.com/docs/en/fast-mode
     if fast_mode:
-        sdk_env["CLAUDE_CODE_FAST_MODE"] = "true"
+        _ensure_fast_mode_in_user_settings()
+        logger.info("[Fast Mode] ACTIVE — will enable user setting source for fastMode")
+        print(
+            "[Fast Mode] ACTIVE — enabling user settings source for CLI to read fastMode"
+        )
+    else:
+        logger.info("[Fast Mode] inactive — not requested for this client")
 
     # Debug: Log git-bash path detection on Windows
     if "CLAUDE_CODE_GIT_BASH_PATH" in sdk_env:
@@ -840,6 +852,12 @@ def create_client(
         # This prevents "File has not been read yet" errors in recovery sessions
         "enable_file_checkpointing": True,
     }
+
+    # Fast mode: enable user setting source so CLI reads fastMode from
+    # ~/.claude/settings.json. Without this, the SDK's default --setting-sources ""
+    # blocks all filesystem settings and the CLI never sees fastMode: true.
+    if fast_mode:
+        options_kwargs["setting_sources"] = ["user"]
 
     # Optional: Allow CLI path override via environment variable
     # The SDK bundles its own CLI, but users can override if needed
--- a/apps/backend/core/fast_mode.py
+++ b/apps/backend/core/fast_mode.py
@@ -8,9 +8,9 @@ per-task fast mode overrides. Shared by both client.py and simple_client.py.
 
 import json
 import logging
-import os
-import tempfile
 from pathlib import Path
+
+from core.file_utils import write_json_atomic
 
 logger = logging.getLogger(__name__)
 
@@ -20,7 +20,7 @@ _fast_mode_atexit_registered = False
 def _write_fast_mode_setting(enabled: bool) -> None:
     """Write fastMode value to ~/.claude/settings.json (atomic read-modify-write).
 
-    Uses write-to-temp-file + atomic rename to prevent corruption when
+    Uses write_json_atomic from core.file_utils to prevent corruption when
     multiple concurrent task processes modify the file simultaneously.
     """
     settings_file = Path.home() / ".claude" / "settings.json"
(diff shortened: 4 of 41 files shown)


AndyMik90 (commented):
## 🤖 Auto Claude PR Review

> **🟠 Needs revision** - 3 issue(s) require attention.

3 issue(s) must be addressed (1 required, 2 recommended), 2 suggestions

### Risk Assessment
| Factor | Level | Notes |
|--------|-------|-------|

### Findings Summary
- **High**: 1 issue(s)
- **Medium**: 2 issue(s)
- **Low**: 2 issue(s)

---

### Findings (5 selected of 5 total)

#### 🟠 [5df2c8a6773f] [HIGH] Fast mode cannot be disabled once enabled via TaskEditDialog
📁 `apps/frontend/src/renderer/components/TaskEditDialog.tsx:251`

When a user toggles fast mode OFF in the task edit dialog, the change does not persist. Line 251 uses `if (fastMode) metadataUpdates.fastMode = true;` which only sets the key when true — it never explicitly sets `fastMode: false`. Since crud-handlers.ts line 475 merges metadata via `{ ...task.metadata, ...updates.metadata }`, the old `fastMode: true` from `task.metadata` is preserved because `updates.metadata` simply omits the key.

Contrast with `requireReviewBeforeCoding` on line 250 which correctly ALWAYS sets the value: `metadataUpdates.requireReviewBeforeCoding = requireReviewBeforeCoding;`. The fastMode line should follow the same pattern.

I verified the merge logic in crud-handlers.ts line 475: `updatedMetadata = { ...task.metadata, ...updates.metadata }` — spread merge preserves keys from `task.metadata` that aren't overridden by `updates.metadata`. | The metadata update logic uses `if (fastMode) metadataUpdates.fastMode = true;` which only sets fastMode when it's true, but never explicitly sets it to false. If a user enables fast mode on a backlog task, then reopens the edit dialog and disables the toggle, the fastMode property is simply omitted from metadataUpdates rather than being set to false. Depending on how `persistUpdateTask` merges metadata (shallow merge vs. replace), the old `fastMode: true` from the existing task_metadata.json may persist. The same pattern exists in TaskCreationWizard (line 459) but is less impactful since creation writes a fresh metadata object. Verified by reading the complete handleSave function (lines 202-257) — there is no `else { metadataUpdates.fastMode = false; }` branch.

**Suggested fix:**
```
Change line 251 to always set the value regardless of toggle state:
`metadataUpdates.fastMode = fastMode;`
This matches the pattern used by `requireReviewBeforeCoding` on line 250.
```

#### 🟡 [2db98dd2a6ba] [MEDIUM] TOCTOU race in concurrent settings file read-modify-write
📁 `apps/backend/core/fast_mode.py:27`

The _write_fast_mode_setting function reads ~/.claude/settings.json, modifies it in memory, then writes it back. While the write itself is atomic (tempfile + os.replace), the read-modify-write cycle is not. When two concurrent task processes both have fast_mode enabled (or one enables while another's atexit disables), the sequence can be: Process A reads {fastMode: true}, Process B reads {fastMode: true}, Process B writes {fastMode: false} (atexit), Process A writes {fastMode: true} — this is fine, but the reverse is problematic: Process A's atexit reads settings, Process B writes fastMode=true, Process A's atexit writes its stale copy back with fastMode=false, clobbering B's change. The docstring acknowledges concurrent modification but atomic rename only prevents corruption, not lost updates. This is a known limitation accepted in the PR description ('CLI subprocess reads settings at startup'), but worth noting as it could cause intermittent fast mode failures when tasks overlap.

**Suggested fix:**
```
Consider using a file lock (e.g., `fcntl.flock` on Unix or a cross-platform lock library) around the read-modify-write cycle. Alternatively, since the CLI reads settings at startup, document that concurrent fast-mode and non-fast-mode tasks may interfere, or use a per-process temp settings file with CLAUDE_CONFIG_DIR isolation instead of a shared global file.
```

#### 🔵 [3dae881f80df] [LOW] Hardcoded validLevels duplicates VALID_THINKING_LEVELS constant
📁 `apps/frontend/src/main/ipc-handlers/settings-handlers.ts:141`

The settings migration code declares `const validLevels = ['low', 'medium', 'high']` as a local array literal, even though VALID_THINKING_LEVELS is already imported from shared/constants (it's used via the sanitizeThinkingLevel import). The import at line 8 already brings in sanitizeThinkingLevel from the same module that exports VALID_THINKING_LEVELS. Using the constant would ensure the migration stays in sync if valid levels change in the future.

**Suggested fix:**
```
Replace `const validLevels = ['low', 'medium', 'high']` with `VALID_THINKING_LEVELS` from the shared constants import (already available since `sanitizeThinkingLevel` is imported from the same module). Update the includes calls to cast appropriately: `(VALID_THINKING_LEVELS as readonly string[]).includes(...)`.
```

#### 🟡 [e38a3aa2288e] [MEDIUM] Reimplements atomic write instead of using core/file_utils.py
📁 `apps/backend/core/fast_mode.py:36`

The _write_fast_mode_setting() function manually implements the atomic write pattern (tempfile.mkstemp + os.fdopen + json.dump + os.replace with cleanup) at lines 36-49. The codebase already has write_json_atomic() in core/file_utils.py which does exactly this, and it is used by 9+ other modules (agents/session.py, spec/validate_pkg/auto_fix.py, runners/roadmap/phases.py, etc.). The existing utility handles the same edge cases (parent dir creation, temp file cleanup on failure) in a well-tested context manager. Searched Grep('write_json_atomic|atomic_write', 'apps/backend') — confirmed 9+ import sites across the backend.

**Suggested fix:**
```
Replace lines 34-49 with:

    settings_file.parent.mkdir(parents=True, exist_ok=True)
    from core.file_utils import write_json_atomic
    write_json_atomic(settings_file, settings)

This reuses the established atomic write utility and reduces duplicated error-handling code.
```

#### 🔵 [6859e0a182d9] [LOW] Underscore-prefixed function exported across module boundaries
📁 `apps/backend/core/fast_mode.py:63`

The function _ensure_fast_mode_in_user_settings() uses a leading underscore (Python private convention) but is imported by core/client.py and core/simple_client.py. Searched Grep('from core\.\w+ import _', 'apps/backend') — this is the ONLY instance in the entire backend where an underscore-prefixed function is imported across modules. All other cross-module imports in core/ use public names (e.g., from core.auth import configure_sdk_authentication, from core.platform import is_windows). Since this function IS the module's public API, it should follow the naming convention of other exported functions.

**Suggested fix:**
```
Rename to ensure_fast_mode_in_user_settings() (no leading underscore) since it is the module's public API. Keep _write_fast_mode_setting() and _disable_fast_mode_on_exit() as private since they are truly internal helpers.
```

---

AndyMik90 (commented):
## ✅ Auto Claude Review - APPROVED

**Status:** Ready to Merge

**Summary:** ## 🟡 Follow-up Review: Merge With Changes

> **🟡 Can merge** - Minor suggestions noted, no blockers.

### Resolution Status
- ✅ **Resolved**: 4 previous findings addressed
- ❌ **Unresolved**: 1 previous findings remain
- 🆕 **New Issues**: 1 new findings in recent changes

### Finding Validation
- 🔍 **Dismissed as False Positives**: 1 findings were re-investigated and found to be incorrect
- ✓ **Confirmed Valid**: 3 findings verified as genuine issues
- 👤 **Needs Human Review**: 1 findings require manual verification

### Verdict
All 20 CI checks passing. 4 of 5 previous findings fully resolved, including the critical HIGH severity fast mode toggle fix. The remaining MEDIUM finding (TOCTOU race in fast_mode.py read-modify-write) was partially addressed - atomic write adopted per recommendation - but the read-modify-write cycle still lacks a file lock. Finding-validator confirmed the race is real but practical impact is very low: only a single boolean toggle with a narrow race window, and worst case is one task missing fast mode activation (not data corruption or security issue). NEW-002 (atexit concern) dismissed as false positive. The only new validated finding is LOW severity (missing sanitize_thinking_level in get_phase_thinking, fully mitigated by downstream fallbacks). Recommend adding file locking or per-process settings isolation in a follow-up PR to fully resolve the TOCTOU race.

### Review Process

---


---

### 💡 Suggestions (1)

*These are non-blocking suggestions for consideration:*

#### 🔵 [NEW-003] [LOW] get_phase_thinking() missing sanitize_thinking_level() for task_metadata values
📁 `apps/backend/phase_config.py:344`

get_phase_thinking() reads thinking levels directly from task_metadata.json and returns them without calling sanitize_thinking_level(). Legacy values like 'ultrathink' could propagate from tasks created before this PR. All downstream numeric consumers (get_thinking_budget, EFFORT_LEVEL_MAP) have fallbacks to 'medium', so this won't cause crashes but it's a defense-in-depth gap in the sanitization strategy this PR establishes.

**Suggested fix:**
```
Apply sanitize_thinking_level() to return values from task_metadata.json:
  return sanitize_thinking_level(phase_thinking.get(phase, DEFAULT_PHASE_THINKING[phase]))
  return sanitize_thinking_level(metadata['thinkingLevel'])
```

---
*This automated review found no blocking issues. The PR can be safely merged.*

*Generated by Auto Claude*

PR #1803, opened 2026-02-12
fix(profiles): complete unified swap infrastructure with API profile support

### LINKED ISSUE #1798: OAuth-to-API Profile Swap Fails: Unified Swap Infrastructure (PR #1794) Built But Never Wired Into Execution Paths
Image

## Summary

When an OAuth Claude profile hits its rate limit or usage threshold, Auto Claude fails to effectively swap to an available API profile (e.g., GLM via z.ai). PR #1794 (commit `28238735` — *"feat(profiles): implement unified profile swapping across OAuth and API accounts"*) built all the unified swap infrastructure (`getBestAvailableUnifiedAccount`, `UnifiedAccount` type, IPC handlers, scoring algorithm) but **never wired it into the actual swap execution paths**. The old OAuth-only code paths still run in all three swap scenarios: proactive, reactive, and spawn-time.

The result is either no swap, a one-way swap that can never reverse, or a **"cosmetic swap"** where the UI shows GLM API but `claude.ai` usage continues on the OAuth account because the process was spawned with OAuth credentials.

**Debug log evidence:**
```
[UsageMonitor:TRACE] Skipping proactive swap for API profile (only supported for OAuth profiles)
```

---

## Current Behavior

1. OAuth profile hits usage threshold (e.g., 95% session) or hard 429 rate limit
2. **Proactive swap** may fire once (OAuth→API direction works via `performProactiveSwap()`) — UI shows swap notification
3. On the **next poll cycle** (30 seconds later), `checkUsageAndSwap()` detects `isAPIProfile = true` and **skips ALL swap logic entirely**, logging the message above
4. System is now stuck on the API profile with no ability to swap back to OAuth when it recovers
5. **Reactive swap** (429 error recovery) calls `getBestAvailableProfile()` which **only searches OAuth profiles** — API profiles are completely invisible as swap targets
6. **"Cosmetic swap" bug**: Even when proactive swap correctly updates `profiles.json`, the process spawn path (`getBestAvailableProfileEnv()`) re-runs OAuth-only selection and **reverses the swap back to the rate-limited OAuth profile**. The new process spawns with OAuth credentials despite the API profile being "active"
7. `claude.ai` continues showing usage on the OAuth account because the actual SDK client uses OAuth credentials

---

## Expected Behavior

1. OAuth profile approaches rate limit → system swaps to best available profile (OAuth **or** API)
2. Running agent processes restart with the **new profile's actual credentials** (API key + base URL for API profiles)
3. If on an API profile, system **can swap back** when OAuth recovers below threshold
4. Reactive swaps (429 recovery at runtime) also consider API profiles as valid targets
5. Auth failures on API profiles (expired key, endpoint down) trigger fallback to working OAuth profiles

---

### Overview: 4 Root Causes Identified

| # | Root Cause | Location | Severity |
|---|-----------|----------|----------|
| 1 | `checkUsageAndSwap()` has `if (!isAPIProfile)` gate — blocks all swap logic once on API profile | `usage-monitor.ts:926` | **Primary** |
| 2 | All 3 reactive swap call sites use OAuth-only `getBestAvailableProfile()` | `agent-process.ts:276,360` + `rate-limit-detector.ts:507` | **Primary** |
| 3 | Spawn-time `getBestAvailableProfileEnv()` reverses proactive swap back to OAuth | `rate-limit-detector.ts:475-593` | **Primary** |
| 4 | PR #1794 unified infrastructure has **zero callers** in actual swap paths — dead code | Multiple files | **Design Gap** |

---

### Root Cause 1: Proactive Swap Gate Blocks API→OAuth Direction

**File:** `apps/frontend/src/main/claude-profile/usage-monitor.ts:926-964`

```typescript
// Step 4: Check thresholds and perform proactive swap (OAuth profiles only)
if (!isAPIProfile) {
    // ... threshold checking and swap logic ...
} else {
    this.debugLog('[UsageMonitor:TRACE] Skipping proactive swap for API profile (only supported for OAuth profiles)');
}
```

**Impact:**
- Once the system swaps TO an API profile, `determineActiveProfile()` (line 1003) returns `isAPIProfile: true`
- ALL subsequent 30-second polling cycles skip the swap logic entirely
- System can **never swap back** from API → OAuth, even when the OAuth profile's rate limit resets
- This produces the exact debug log message reported by users

**Additionally**, `handleAuthFailure()` at line 1171 also blocks API profiles:

```typescript
if (isAPIProfile || !settings.enabled || !settings.proactiveSwapEnabled) {
    this.debugLog('[UsageMonitor] Auth failure detected but proactive swap is disabled or using API profile, skipping swap');
    return;
}
```

If an API key expires or the endpoint goes down, the system is stuck with **no automatic recovery**.

---

### Root Cause 2: Reactive Swap Only Considers OAuth Profiles

Three separate call sites all use the old OAuth-only `getBestAvailableProfile()`:

| Call Site | File:Line | Function | What It Does |
|-----------|-----------|----------|--------------|
| Rate limit auto-swap | `agent-process.ts:276` | `handleRateLimitWithAutoSwap()` | Hit 429 during execution → find alternative |
| Auth failure auto-swap | `agent-process.ts:360` | Same handler, auth failure path | Auth error → find alternative |
| Spawn-time swap | `rate-limit-detector.ts:507` | `getBestAvailableProfileEnv()` | New process spawn → pick best profile |

All three call:
```typescript
const bestProfile = profileManager.getBestAvailableProfile(currentProfileId);
```

`ClaudeProfileManager.getBestAvailableProfile()` at `claude-profile-manager.ts:667-671`:
```typescript
getBestAvailableProfile(excludeProfileId?: string): ClaudeProfile | null {
    const settings = this.getAutoSwitchSettings();
    const priorityOrder = this.getAccountPriorityOrder();
    return getBestAvailableProfile(this.data.profiles, settings, excludeProfileId, priorityOrder);
}
```

This only passes `this.data.profiles` (OAuth profiles). **API profiles are never considered.** So when a running agent hits a 429 and the only fallback is an API profile, the reactive swap finds nothing and fails silently.

---

### Root Cause 3: Spawn-Time Swap Reversal (Explains "Cosmetic Swap")

This is the mechanism that explains why `claude.ai` still shows usage after an apparent swap to GLM API:

**Sequence:**
1. `performProactiveSwap()` correctly sets API profile in `profiles.json` via `setActiveAPIProfile()` ✓
2. UI shows swap notification ✓
3. `OperationRegistry.restartOperationsOnProfile()` kills and respawns agent processes
4. New process spawns → `setupProcessEnvironment()` runs
5. `getBestAvailableProfileEnv()` in `rate-limit-detector.ts` runs the **OAuth-only** `getBestAvailableProfile()` logic
6. This function **re-selects the rate-limited OAuth profile** (or another OAuth profile) because it doesn't know API profiles exist
7. Process environment gets **both** `CLAUDE_CONFIG_DIR` (OAuth) **and** `ANTHROPIC_AUTH_TOKEN`/`ANTHROPIC_BASE_URL` (API) set
8. The SDK client resolves credentials from `CLAUDE_CONFIG_DIR` first → reads OAuth token from OS keychain → API calls go through Claude OAuth

**Result:** The swap is real at the config level but **reversed at process spawn time**. The UI shows the API profile as active, but the actual SDK client uses OAuth credentials.

---

### Root Cause 4: PR #1794 Unified Infrastructure Is Dead Code

Commit `28238735` added the following, but **none of it is called by any swap execution path**:

| What Was Built | File | Callers in Swap Paths |
|---------------|------|----------------------|
| `getBestAvailableUnifiedAccount()` scoring function | `profile-scorer.ts` | **0** |
| `getBestAvailableUnifiedAccount()` manager method | `claude-profile-manager.ts` | **0** |
| `QUEUE_GET_BEST_UNIFIED_ACCOUNT` IPC handler | `queue-routing-handlers.ts:103` | **0** (only `browser-mock.ts`) |
| `getBestUnifiedAccount` preload API | `queue-api.ts` | **0** (only `browser-mock.ts`) |
| `UnifiedAccount` type + conversion utils | `shared/types/` + `shared/utils/` | Used by scorer, but scorer never called |
| `scoreUnifiedAccount()` | `profile-scorer.ts` | **0** from swap paths |

**Grep confirmation:** `getBestAvailableUnifiedAccount` appears in:
- `profile-scorer.ts` (definition)
- `claude-profile-manager.ts` (wrapper method)
- `queue-routing-handlers.ts` (IPC handler — never called from renderer)
- `browser-mock.ts` (test stub)
- **Zero** appearances in `usage-monitor.ts`, `agent-process.ts`, `rate-limit-detector.ts`, or any renderer component

Meanwhile, `performProactiveSwap()` in `usage-monitor.ts:1887-1923` has its **own inline duplicate** of unified list-building logic rather than using the new scorer. The only change to `usage-monitor.ts` in the commit (7 lines) was a variable rename from `bestAccount.id` to `rawProfileId` — **zero functional changes to swap logic**.

The commit message itself acknowledges this: *"existing getBestAvailableProfile continues to work for OAuth-only scenarios"* — but **ALL** actual swap scenarios still use the OAuth-only path.

---

## Impact

- **Who is affected:** Any user with a mixed OAuth + API profile setup who relies on automatic profile swapping
- **Severity:** High — autonomous tasks stall when the primary OAuth profile is rate-limited; the user's paid API profile (GLM/z.ai) sits completely idle despite being configured as a fallback
- **User confusion:** The UI may show a successful swap to GLM API, but actual API calls continue through Claude OAuth — users see contradictory signals
- **No workaround via settings:** Even configuring account switching limits doesn't fix this because the reactive swap path is fundamentally OAuth-only
- **Manual workaround:** Users must notice the rate limit and manually switch profiles via the UI

---

## Steps to Reproduce

1. Set up an OAuth Claude profile (e.g., Claude Max subscription) as the primary profile
2. Set up a GLM API profile (e.g., via z.ai endpoint) as a secondary profile with valid API key
3. Enable auto-switch in Settings → Account Switching
4. Enable proactive swap with threshold (e.g., 95% session usage)
5. Configure account priority order: OAuth first, API second
6. Run autonomous tasks until the OAuth profile approaches the session threshold

**Observe (Proactive Path):**

7. The proactive swap fires and UI shows "Swapped to GLM API" notification
8. Wait 30 seconds for the next usage check cycle
9. Check debug logs: `"[UsageMonitor:TRACE] Skipping proactive swap for API profile (only supported for OAuth profiles)"` appears
10. Check `claude.ai` → usage is still ticking up on the OAuth account (not using GLM)
11. The system never swaps back to OAuth even after the rate limit resets

**Observe (Reactive Path):**

12. If the OAuth profile gets a hard 429 during agent execution, the reactive swap in `agent-process.ts` only looks for other OAuth profiles
13. If no other OAuth profile exists, the swap fails silently
14. The GLM API profile is completely ignored despite being available and configured

---

### Change 1: Remove `isAPIProfile` gate in `checkUsageAndSwap()`

**File:** `apps/frontend/src/main/claude-profile/usage-monitor.ts`
**Lines:** 926-964

**What:** Remove or relax the `if (!isAPIProfile)` guard. For API profiles with usage monitoring support (z.ai, ZHIPU providers that return usage data), check thresholds normally. For API profiles without usage data, periodically check if any OAuth profile has recovered below its threshold and swap back.

**Why:** This is the primary blocker. `performProactiveSwap()` already handles both OAuth and API targets at lines 1906-1923 — it just needs to be invoked regardless of current profile type.

### Change 2: Remove `isAPIProfile` guard in `handleAuthFailure()`

**File:** `apps/frontend/src/main/claude-profile/usage-monitor.ts`
**Line:** 1171

**What:** Remove `isAPIProfile` from the early return condition:
```typescript
// Before:
if (isAPIProfile || !settings.enabled || !settings.proactiveSwapEnabled) { return; }

// After:
if (!settings.enabled || !settings.proactiveSwapEnabled) { return; }
```

**Why:** If an API key expires or endpoint goes down, the system should fall back to a working OAuth profile.

### Change 3: Wire unified account selection into reactive swap

**File:** `apps/frontend/src/main/agent/agent-process.ts`
**Lines:** 276, 360

**What:** Replace `profileManager.getBestAvailableProfile(currentProfileId)` with `await profileManager.getBestAvailableUnifiedAccount(currentUnifiedId)`. Use `extractProfileId()` and `isAPIAccountId()` from `shared/utils/unified-account.ts` to handle the `UnifiedAccount` return type and apply the correct profile (OAuth vs API).

**Why:** Ensures 429 recovery and auth failure recovery can target API profiles when no other OAuth profiles are available.

### Change 4: Wire unified selection into spawn-time swap

**File:** `apps/frontend/src/main/rate-limit-detector.ts`
**Lines:** 475-593 (specifically around line 507)

**What:** Replace the OAuth-only `getBestAvailableProfile()` logic in `getBestAvailableProfileEnv()` with unified account selection that respects the currently active API profile. When an API profile is active, do NOT override it with an OAuth profile.

**Why:** This is the fix for the "cosmetic swap" bug. Currently, this function actively reverses a valid proactive swap by re-selecting OAuth at spawn time.

### Change 5: Consolidate `performProactiveSwap()` to use the unified scorer

**File:** `apps/frontend/src/main/claude-profile/usage-monitor.ts`
**Lines:** 1887-1923

**What:** Replace the inline unified list-building logic with a call to `profileManager.getBestAvailableUnifiedAccount()`. This eliminates the duplicated logic and ensures a single source of truth for scoring.

**Why:** Currently there are two separate implementations of unified account selection — the proper one in `profile-scorer.ts` (built in PR #1794 but unused) and the ad-hoc one inline in `performProactiveSwap()`. Consolidating ensures consistent scoring, including threshold enforcement.

### Change 6: Expose sync-compatible unified selection

**File:** `apps/frontend/src/main/claude-profile-manager.ts`

**What:** Ensure `getBestAvailableUnifiedAccount()` is accessible in both async and sync contexts, or refactor `getBestAvailableProfileEnv()` to be async. Some callers (like `setupProcessEnvironment`) may need a synchronous variant.

**Why:** The current `getBestAvailableUnifiedAccount()` is async (loads API profiles from disk). The spawn-time swap path may need a cached/sync version.

---

### Files Requiring Changes (Summary)

| File | Change | Priority |
|------|--------|----------|
| `usage-monitor.ts:926` | Remove `if (!isAPIProfile)` gate | Critical |
| `usage-monitor.ts:1171` | Remove `isAPIProfile` from `handleAuthFailure()` | High |
| `agent-process.ts:276,360` | Use `getBestAvailableUnifiedAccount()` in reactive swap | Critical |
| `rate-limit-detector.ts:507` | Use unified selection in spawn-time swap | Critical |
| `usage-monitor.ts:1887-1923` | Consolidate to use `getBestAvailableUnifiedAccount()` | Medium |
| `claude-profile-manager.ts` | Expose sync-compatible unified method | Medium |

---

### Unit Tests
- [ ] `checkUsageAndSwap()` triggers proactive swap when API profile is active and an OAuth profile has recovered
- [ ] `handleRateLimitWithAutoSwap()` selects an API profile as swap target when no other OAuth is available
- [ ] `handleAuthFailure()` swaps away from API profile when auth fails
- [ ] `getBestAvailableProfileEnv()` respects active API profile and does not reverse to OAuth
- [ ] `scoreUnifiedAccount()` correctly scores both OAuth and API profiles

### Integration Tests
- [ ] OAuth at 95% usage → verify swap to API profile → verify API credentials reach agent process (not OAuth)
- [ ] 429 during agent execution with only API profile available → verify reactive swap targets API
- [ ] API→OAuth swap-back when OAuth rate limit resets
- [ ] API key expires → verify fallback to OAuth profile

### Edge Cases
- [ ] Swap loop prevention — add cooldown between swaps or max-swap-count per time window
- [ ] Both OAuth and API profiles exhausted — graceful degradation
- [ ] Multiple API profiles available — correct scoring/priority

---

## Technical Notes

- `performProactiveSwap()` already handles both OAuth and API targets correctly (lines 1906-1923: uses `setActiveProfile()` for OAuth, `setActiveAPIProfile()` for API). The fix is primarily about removing gates that prevent invocation and wiring the unified scorer.
- The `OperationRegistry` restart mechanism is sound — it correctly kills and respawns processes after swap. The restart infrastructure works; only the swap trigger and target selection logic needs fixing.
- API profiles store credentials in `profiles.json` (`apiKey` field); OAuth profiles store tokens in OS Keychain. The credential systems are completely separate, which is why the swap must correctly switch between `CLAUDE_CONFIG_DIR` (OAuth) and `ANTHROPIC_AUTH_TOKEN`/`ANTHROPIC_BASE_URL` (API) env vars.
- `getOAuthModeClearVars()` in `env-utils.ts` correctly clears stale `ANTHROPIC_*` vars when no API profile is active — this logic is sound.
- Risk: Allowing swap FROM API profiles could create swap loops if thresholds aren't configured properly. Consider adding a cooldown between swaps (e.g., 60 seconds minimum between profile switches) or a maximum swap count per session.
- This is a **design gap** (feature partially implemented in PR #1794), not a regression. The forward direction (OAuth→API) was partially built, but reverse direction and reactive paths were never updated.

---

## Related

- PR #1794 — `feat(profiles): implement unified profile swapping across OAuth and API accounts` (the commit that built the infrastructure)
- Commit `28238735` — the specific commit
- `apps/frontend/src/main/claude-profile/usage-monitor.ts` — primary file
- `apps/frontend/src/main/agent/agent-process.ts` — reactive swap
- `apps/frontend/src/main/claude-profile/profile-scorer.ts` — unified scorer (currently dead code)

VDT-91:
needs testing. going to bed

Fixes #1798


AndyMik90 (changes requested):
## 🤖 Auto Claude PR Review

### Merge Verdict: 🔴 BLOCKED

> **🔴 Blocked** - Merge conflicts must be resolved before merge.

Blocked: PR has merge conflicts with base branch. Resolve conflicts before merge.

### Risk Assessment
| Factor | Level | Notes |
|--------|-------|-------|
| Security Impact | Low | Based on security findings |

- Merge Conflicts: PR has conflicts with base branch that must be resolved

### Findings Summary
- **High**: 1 issue(s)
- **Medium**: 6 issue(s)
- **Low**: 1 issue(s)

---

#### 🟡 [11642356e6a9] [MEDIUM] API credentials visible in shell history via inline command
📁 `apps/frontend/src/main/terminal/claude-integration-handler.ts:316`

The new `api-profile` case in `buildClaudeShellCommand()` embeds API keys directly in the shell command string (e.g., `ANTHROPIC_AUTH_TOKEN='sk-ant-xxx' bash -c "exec claude"`). Unlike the `temp-file` method used for OAuth tokens (which writes credentials to a 0600-permission temp file, sources it, then deletes it), the api-profile method puts the API key directly in the command text. The parent shell running in the PTY terminal records this full command in its history file (~/.bash_history or ~/.zsh_history). While `HISTFILE=` is set, it only applies to the child `bash -c` subshell — it does NOT prevent the parent shell from recording the command. Compare with the temp-file method at lines 284-288 which was specifically designed to keep credentials out of the command string. The existing temp-file pattern should be reused for API keys to maintain equivalent credential protection.

**Suggested fix:**
```
Use the same temp-file pattern as OAuth tokens: write API env vars to a temporary file with 0600 permissions, source it, delete it, then exec Claude. This keeps credentials out of the command string and shell history. For example, generate a temp file containing `export ANTHROPIC_AUTH_TOKEN=... export ANTHROPIC_BASE_URL=...` and use the existing temp-file code path.
```

#### 🟡 [f43b4bf164cd] [MEDIUM] API credentials logged to console output in debug mode
📁 `apps/frontend/src/main/terminal/claude-integration-handler.ts:1189`

When DEBUG=true, the full shell command (including ANTHROPIC_AUTH_TOKEN with the actual API key) is logged via `debugLog()` which writes to `console.warn`. This occurs at two call sites for the api-profile method. If a user enables debug mode for troubleshooting and shares their logs (e.g., in a bug report or support request), the API key is exposed. The debugLog function at debug-logger.ts:13-16 writes directly to console.warn with no redaction. The existing OAuth token paths avoid this because the token is in a separate temp file, not in the command string.

**Suggested fix:**
```
Redact sensitive environment variable values before logging. For example, mask the ANTHROPIC_AUTH_TOKEN value (show only first 8 and last 4 characters) in the logged command, or log 'Executing command (api-profile method): [command with N env vars redacted]' instead of the full command string.
```

#### 🔵 [ed1d3ade0922] [LOW] [NEEDS REVIEW] Env var key name not shell-escaped in Unix api-profile command
📁 `apps/frontend/src/main/terminal/claude-integration-handler.ts:317`

In the `api-profile` case of `buildClaudeShellCommand()`, the environment variable key name is interpolated directly into the shell command without escaping on Unix: `${key}=${escapeShellArg(value)}`. While the value is properly escaped via `escapeShellArg()`, the key name is not. The function signature accepts `envVars: Record<string, string>` which permits arbitrary key names. Currently all callers use hardcoded ANTHROPIC_* key names (safe), but the function contract doesn't enforce this. A future caller passing user-controlled keys could enable command injection. On Windows (line 310), both key and value are escaped via `escapeForWindowsDoubleQuote()`. Searched callers with Grep: invokeClaude (line 1186) uses hardcoded keys; invokeClaudeAsync (line 1419) uses getAPIProfileEnv() which also uses hardcoded keys.

**Suggested fix:**
```
Validate that key names match a safe pattern (e.g., /^[A-Z_][A-Z0-9_]*$/) before interpolation, or escape the key name. This provides defense-in-depth against future callers passing unsanitized keys.
```

#### 🟡 [08134f297562] [MEDIUM] Non-null assertion on potentially null value (oauthProfile!)
📁 `apps/frontend/src/main/agent/agent-process.ts:389`

In `handleAuthFailureWithAutoSwap()`, `profileManager.getProfile(bestAccount.id)` may return `null` or `undefined` (e.g., if the profile was deleted between selection and verification). Line 389 uses `oauthProfile!` (non-null assertion) as an argument to `isProfileAuthenticated()`, which would crash with a runtime TypeError if `oauthProfile` is actually null. The guard on line 389 only checks `oauthProfile?.isAuthenticated` with optional chaining, but then passes `oauthProfile!` to the second function which will dereference it. If `getProfile()` returns null AND `isAuthenticated` is falsy (which it would be), the `!oauthProfile?.isAuthenticated` is `true`, so code enters the condition and `isProfileAuthenticated(oauthProfile!)` is never reached. However, this reasoning is fragile — if `getProfile` returns an object where `isAuthenticated` is truthy but `isProfileAuthenticated()` returns false, the non-null assertion is safe. The pattern is still dangerous and unclear. A simple null guard (`if (!oauthProfile)`) before the check would be cleaner and safer. | In `handleAuthFailureWithAutoSwap`, `getProfile(bestAccount.id)` returns `ClaudeProfile | undefined`, but line 389 uses the non-null assertion operator `oauthProfile!` when passing to `isProfileAuthenticated()`. If the OAuth profile was recently deleted or the bestAccount.id doesn't match any stored profile, `oauthProfile` will be `undefined` and the non-null assertion will pass `undefined` to `isProfileAuthenticated()`, which may throw a TypeError accessing properties on undefined.

The condition `!oauthProfile?.isAuthenticated` correctly guards the first part with optional chaining, but the second operand `!profileManager.isProfileAuthenticated(oauthProfile!)` bypasses the null check via `!` assertion.

**Suggested fix:**
```
Add explicit null check: `if (!oauthProfile || (!oauthProfile.isAuthenticated && !profileManager.isProfileAuthenticated(oauthProfile))) {`
```

#### 🟠 [b099660ccb86] [HIGH] Profile swap logic duplicated across 3 files (5 occurrences)
📁 `apps/frontend/src/main/agent/agent-process.ts:294`

The identical pattern of `if (bestAccount.type === 'oauth') { profileManager.setActiveProfile(id); await setActiveAPIProfile(null); } else { await setActiveAPIProfile(id); }` is duplicated across 3 files with 5 occurrences: agent-process.ts (lines 294-306, 396-408), rate-limit-detector.ts (lines 540-556), and usage-monitor.ts (lines 1916-1936). Each copy includes the same dynamic import, try/catch, and error logging. This is a classic maintainability issue — if the swap logic changes (e.g., adding a new profile type, or needing to emit an event), all 5 locations must be updated. This should be extracted to a shared utility function like `switchToUnifiedAccount(profileManager, bestAccount)` in a common module.

**Suggested fix:**
```
Extract to a shared function in a common module (e.g., profile-swap-utils.ts): `export async function switchToUnifiedAccount(profileManager, bestAccount) { ... }` and call it from all 3 files.
```

#### 🟡 [7233b605b828] [MEDIUM] Missing error handling for setActiveAPIProfile in non-OAuth branch
📁 `apps/frontend/src/main/agent/agent-process.ts:303`

In `handleRateLimitWithAutoSwap()`, when `bestAccount.type` is not 'oauth' (i.e., it's an API profile), the `setActiveAPIProfile()` call on line 304-305 is NOT wrapped in a try/catch, unlike the OAuth branch (lines 297-302) and unlike the same pattern in `handleAuthFailureWithAutoSwap()` (lines 405-408) and `usage-monitor.ts` (lines 1929-1935). If the dynamic import or `setActiveAPIProfile()` throws, the error will propagate up as an unhandled rejection, potentially crashing the swap flow. This inconsistency within the same file is a clear oversight.

**Suggested fix:**
```
Wrap in try/catch like the OAuth branch: `try { const { setActiveAPIProfile } = await import('...'); await setActiveAPIProfile(bestAccount.id); } catch (error) { console.error('[AgentProcess] Failed to set active API profile:', error); }`
```

#### 🟡 [75b8eef620fe] [MEDIUM] getBestAvailableUnifiedAccount filters out unavailable OAuth profiles with no fallback
📁 `apps/frontend/src/main/claude-profile-manager.ts:716`

Unlike `getBestAvailableProfile()` (line 380-465 in profile-scorer.ts) which includes ALL profiles in scoring and returns a 'least bad' option when none pass availability checks, `getBestAvailableUnifiedAccount()` at line 721 does `if (!availability.available) continue;` which completely removes unavailable OAuth profiles from consideration.

This means when ALL OAuth profiles are rate-limited and ALL API profiles are excluded, this method returns `null` — while `getBestAvailableProfile()` would return the 'least bad' OAuth profile (e.g., one whose rate limit resets soonest). This asymmetry could cause the proactive swap in usage-monitor.ts (line 1890) to emit 'no_alternative' when there IS a usable-soon OAuth profile, blocking the swap entirely instead of using the 'least bad' option.

Note: The standalone `getBestAvailableUnifiedAccount()` in profile-scorer.ts (line 251-363) correctly includes all profiles and returns 'least bad' options. The ClaudeProfileManager wrapper method does not delegate to it and instead has its own filtering logic that discards unavailable profiles.

**Suggested fix:**
```
Consider delegating to the standalone `getBestAvailableUnifiedAccount()` from profile-scorer.ts which already implements the full scoring + 'least bad' fallback algorithm. Alternatively, include unavailable OAuth profiles in the unifiedAccounts list with a lower priority score so they serve as fallback candidates.
```

#### 🟡 [d7eb2a331368] [MEDIUM] Duplicate type: UnifiedSwapTarget duplicates shared UnifiedAccount
📁 `apps/frontend/src/main/claude-profile-manager.ts:60`

The PR introduces `UnifiedSwapTarget` in `claude-profile-manager.ts` (lines 60-65) with fields `{id, name, type, priorityIndex}` while the shared type `UnifiedAccount` in `shared/types/unified-account.ts` already serves the same purpose with a richer interface. This creates a type inconsistency at the IPC boundary: the queue-routing-handlers.ts handler (line 109) returns `UnifiedSwapTarget`, but the preload `queue-api.ts` (line 85) declares the return type as `UnifiedAccount`. The profile-scorer.ts `getBestAvailableUnifiedAccount()` function (line 256) correctly returns `UnifiedAccount`, but `ClaudeProfileManager.getBestAvailableUnifiedAccount()` (line 707) returns `UnifiedSwapTarget`. This means the renderer receives a `UnifiedSwapTarget` (4 fields) but the preload API type expects `UnifiedAccount` (15+ fields) — a silent shape mismatch.

Searched `Grep('UnifiedSwapTarget|UnifiedAccount', '.')` — confirmed both types exist and are used on opposite sides of the IPC boundary. The shared `UnifiedAccount` type is the canonical one (defined in shared/types/, re-exported from shared/types/index.ts). The main process `getBestAvailableUnifiedAccount` in profile-scorer.ts already returns `UnifiedAccount`, so the manager's parallel implementation should align.

**Suggested fix:**
```
Remove the `UnifiedSwapTarget` interface and use the shared `UnifiedAccount` type throughout. Update `ClaudeProfileManager.getBestAvailableUnifiedAccount()` to delegate to `profile-scorer.ts`'s `getBestAvailableUnifiedAccount()` which already returns `UnifiedAccount`, or convert the result to `UnifiedAccount` before returning.
```

---

PR #1887, opened 2026-02-22
fix: suppress OAuth re-auth UI for API profiles

alexx-ftw:
## Summary
- gate UsageIndicator re-auth UI to OAuth mode only by checking whether an API profile is active
- prevent Custom Endpoint mode from showing OAuth re-auth warnings when usage is unavailable
- add regression tests for API mode (no re-auth UI) and OAuth mode (re-auth UI shown)

## Test plan
- [x] npm --prefix \"apps/frontend\" run typecheck
- [x] npm --prefix \"apps/frontend\" run test -- src/renderer/components/UsageIndicator.test.tsx src/renderer/components/AuthStatusIndicator.test.tsx --run

## Summary by CodeRabbit

* **Tests**
  * Added Vitest tests for the usage indicator covering API vs OAuth profiles, re-auth flows, and stable translations; updated test import resolution to point to renderer sources.

* **Bug Fixes**
  * Re-authentication messaging now appears only for OAuth profiles.
  * Usage availability UI correctly reflects active profile (API vs OAuth) and avoids false re-auth prompts.

* **New Features**
  * Renderer now receives real-time task status change notifications so the UI stays in sync with task recovery and status updates.

### DIFF
--- a/apps/backend/core/auth.py
+++ b/apps/backend/core/auth.py
@@ -965,6 +965,14 @@ def get_sdk_env_vars() -> dict[str, str]:
         if bash_path:
             env["CLAUDE_CODE_GIT_BASH_PATH"] = bash_path
 
+    # Explicitly unset CLAUDECODE in SDK subprocess environment.
+    #
+    # Claude Code CLI refuses to start when this marker env var is present,
+    # treating it as a nested Claude Code session. Auto-Claude can run inside
+    # a Claude Code terminal during debugging, so we must clear this marker for
+    # child SDK/CLI processes.
+    env["CLAUDECODE"] = ""
+
     # Explicitly unset PYTHONPATH in SDK subprocess environment to prevent
     # pollution of agent subprocess environments. This fixes ACS-251 where
     # external projects with different Python versions would fail due to
--- a/apps/backend/core/client.py
+++ b/apps/backend/core/client.py
@@ -18,6 +18,7 @@ import logging
 import os
 import threading
 import time
+from collections.abc import Callable
 from pathlib import Path
 from typing import Any
 
@@ -536,6 +537,7 @@ def create_client(
     betas: list[str] | None = None,
     effort_level: str | None = None,
     fast_mode: bool = False,
+    stderr_callback: Callable[[str], None] | None = None,
 ) -> ClaudeSDKClient:
     """
     Create a Claude Agent SDK client with multi-layered security.
@@ -571,6 +573,8 @@ def create_client(
                   the "user" setting source so the CLI reads fastMode from
                   ~/.claude/settings.json. Requires extra usage enabled on Claude
                   subscription; falls back to standard speed automatically.
+        stderr_callback: Optional callback invoked with each stderr line emitted
+                        by the Claude CLI process.
 
     Returns:
         Configured ClaudeSDKClient
@@ -958,6 +962,9 @@ def create_client(
         # This prevents "File has not been read yet" errors in recovery sessions
         "enable_file_checkpointing": True,
     }
+
+    if stderr_callback:
+        options_kwargs["stderr"] = stderr_callback
 
     # Fast mode: enable user setting source so CLI reads fastMode from
     # ~/.claude/settings.json. Without this, the SDK's default --setting-sources ""
--- a/apps/backend/prompts/spec_quick.md
+++ b/apps/backend/prompts/spec_quick.md
@@ -31,46 +31,46 @@ That's it. No deep analysis needed.
 
 ## PHASE 2: CREATE MINIMAL SPEC
 
-Create a concise `spec.md`:
+Create a concise `spec.md` that includes the required validator sections:
 
 ```bash
 cat > spec.md << 'EOF'
-# Quick Spec: [Task Name]
+# Specification: [Task Name]
 
-## Task
-[One sentence description]
+## Overview
+[One paragraph describing what is being changed and why]
 
-## Files to Modify
+## Workflow Type
+**Type**: simple
+
+## Task Scope
+### Files to Modify
 - `[path/to/file]` - [what to change]
 
-## Change Details
+### Change Details
 [Brief description of the change - a few sentences max]
 
-## Verification
+## Success Criteria
 - [ ] [How to verify the change works]
-
-## Notes
-[Any gotchas or considerations - optional]
 EOF
 ```
 
-**Keep it short!** A simple spec should be 20-50 lines, not 200+.
+**Keep it short!** A simple spec should be concise and focused.
 
 ---
 
 ## PHASE 3: CREATE SIMPLE PLAN
 
-Create `implementation_plan.json`:
+Create `implementation_plan.json` with current schema fields:
 
 ```bash
 cat > implementation_plan.json << 'EOF'
 {
-  "spec_name": "[spec-name]",
+  "feature": "[task description]",
   "workflow_type": "simple",
-  "total_phases": 1,
-  "recommended_workers": 1,
   "phases": [
     {
+      "id": "phase-1",
       "phase": 1,
       "name": "Implementation",
       "description": "[task description]",
@@ -86,17 +86,19 @@ cat > implementation_plan.json << 'EOF'
           "patterns_from": [],
           "verification": {
             "type": "manual",
-            "run": "[verification step]"
+            "instructions": "[verification step]"
           }
         }
       ]
     }
   ],
-  "metadata": {
-    "created_at": "[timestamp]",
-    "complexity": "simple",
-    "estimated_sessions": 1
-  }
+  "summary": {
+    "total_phases": 1,
+    "total_subtasks": 1,
+    "recommended_workers": 1
+  },
+  "created_at": "[timestamp]",
+  "updated_at": "[timestamp]"
 }
 EOF
 ```
@@ -146,18 +148,22 @@ Ready for implementation.
 
 **spec.md**:
 ```markdown
-# Quick Spec: Button Color Change
+# Specification: Button Color Change
 
-## Task
+## Overview
 Update primary button color from blue (#3B82F6) to green (#22C55E).
 
-## Files to Modify
+## Workflow Type
+**Type**: simple
+
+## Task Scope
+### Files to Modify
 - `src/components/Button.tsx` - Update color constant
 
-## Change Details
+### Change Details
 Change the `primaryColor` variable from `#3B82F6` to `#22C55E`.
 
-## Verification
+## Success Criteria
 - [ ] Buttons appear green in the UI
 - [ ] No console errors
 ```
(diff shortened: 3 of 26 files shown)


AndyMik90 (changes requested):
## 🤖 Auto Claude PR Review

PR #1887 reviewed: 14 findings (3 structural issues). Verdict: needs_review.

### Findings (14 selected of 14 total)

#### 🟠 [QLT-1] [HIGH] Vitest alias change may break other tests
📁 `apps/frontend/vitest.config.ts:27`

Changing the `@` alias from `src` to `src/renderer` is a global config change that will break any existing test importing main-process or shared code via `@/`. For example, `@/shared/utils/...` or `@/main/...` imports would now resolve to `src/renderer/shared/utils/...` which doesn't exist. This change appears to have been made to satisfy a single test file's import (`@/stores/settings-store`) rather than fixing the import in that test.

**Suggested fix:**
```
Revert the alias to `'@': resolve(__dirname, 'src')` and update the test import to use `@renderer/stores/settings-store` or `@/renderer/stores/settings-store` instead.
```

#### 🟡 [QLT-2] [MEDIUM] Stale closure risk in useEffect with partial dependency array
📁 `apps/frontend/src/renderer/components/UsageIndicator.tsx:366`

The useEffect sets up IPC listeners that capture `isUsingApiProfile` in their closures, but the dependency array only includes `[isUsingApiProfile]`. When `isUsingApiProfile` changes, the effect re-runs, tearing down old listeners and creating new ones. However, the `requestAllProfilesUsage` fetch inside the effect also re-fires on every toggle, which may cause unnecessary network requests. More critically, the `onUsageUpdated` listener callback doesn't use `isUsingApiProfile` but gets torn down and recreated anyway, potentially losing in-flight events.

**Suggested fix:**
```
Consider splitting into separate useEffects: one for the IPC listeners (stable, runs once) and one that reacts to `isUsingApiProfile` changes to recompute derived state, or use a ref for `isUsingApiProfile` inside the listener callbacks.
```

#### 🟡 [QLT-3] [MEDIUM] Duplicated profile filtering and re-auth logic
📁 `apps/frontend/src/renderer/components/UsageIndicator.tsx:329`

The logic for filtering non-active profiles, finding the active profile, and computing `setActiveProfileNeedsReauth(!isUsingApiProfile && Boolean(activeProfile?.needsReauthentication))` is duplicated in three places: the `onAllProfilesUsageUpdated` callback, the initial `requestAllProfilesUsage` promise handler, and conceptually in the derived `showUnavailableReauth`/`showReauthForActiveAccount` variables. This violates DRY and increases the risk of inconsistency.

**Suggested fix:**
```
Extract a helper function like `processAllProfilesData(data, isUsingApiProfile)` that returns `{ otherProfiles, activeProfileNeedsReauth }` and call it from both the callback and the initial fetch handler.
```

#### 🟡 [QLT-4] [MEDIUM] Redundant re-auth suppression at multiple layers
📁 `apps/frontend/src/renderer/components/UsageIndicator.tsx:314`

Re-auth is suppressed both when setting state (`setActiveProfileNeedsReauth(!isUsingApiProfile && ...)`) AND when reading it (`showUnavailableReauth = !isUsingApiProfile && activeProfileNeedsReauth`). The double-gating is unnecessary and confusing — if the state is already guaranteed to be false for API profiles, the derived variable doesn't need the check, or vice versa. This creates a maintenance trap where future developers may only update one layer.

**Suggested fix:**
```
Choose one place to enforce the `isUsingApiProfile` check: either in the state setter or in the derived variables, not both.
```

#### 🔵 [QLT-5] [LOW] Naming convention: `isUsingApiProfile` could be misleading
📁 `apps/frontend/src/renderer/components/UsageIndicator.tsx:85`

The variable `isUsingApiProfile` is derived from `Boolean(activeProfileId)`. The name suggests it specifically detects API profile mode, but `activeProfileId` being non-null may not exclusively mean 'API profile' — it depends on the store semantics. If OAuth profiles can also have an `activeProfileId`, this logic would be incorrect.

**Suggested fix:**
```
Verify that `activeProfileId` being truthy always means 'API profile mode'. If so, add a comment clarifying this invariant. If not, use a more specific check like `activeProfile?.type === 'api'`.
```

#### 🔵 [QLT-6] [LOW] Test uses tightly-coupled type derived from helper function
📁 `apps/frontend/src/renderer/components/UsageIndicator.test.tsx:54`

The test defines `AllProfilesUsagePayload` as `ReturnType<typeof buildAllProfilesUsageResponse>['data']`, coupling the type to the test helper's implementation. If the helper shape changes, the type silently changes too. This is fragile and makes the test harder to understand.

**Suggested fix:**
```
Import or define the actual payload type from the source code (e.g., from shared types) rather than deriving it from a test helper.
```

#### 🔵 [QLT-7] [LOW] Missing test cleanup of window.electronAPI
📁 `apps/frontend/src/renderer/components/UsageIndicator.test.tsx:96`

The test sets `(window as any).electronAPI` in `beforeEach` but never cleans it up in an `afterEach`. This can leak state between test suites and cause hard-to-debug failures when tests run in a different order.

**Suggested fix:**
```
Add an `afterEach` block that deletes or resets `(window as any).electronAPI`.
```

#### 🔵 [QLT-8] [LOW] sendStatusChange helper not called in all early-return paths
📁 `apps/frontend/src/main/ipc-handlers/task/execution-handlers.ts:977`

The `sendStatusChange` helper is added to some early-return recovery paths but there may be other early returns in the handler (e.g., the initial `if (!task)` check) where status notification is also missing. A systematic review of all early returns would be warranted to ensure consistency.

**Suggested fix:**
```
Audit all early-return paths in the handler to determine if `sendStatusChange` should be called, and document intentional omissions.
```

#### 🟠 [DEEP-1] [HIGH] Vitest alias change from `@` → `src/renderer` breaks all non-renderer imports project-wide
📁 `apps/frontend/vitest.config.ts:27`

The alias `@` was changed from `resolve(__dirname, 'src')` to `resolve(__dirname, 'src/renderer')`. This means any existing test that imports from `@/main/...` or `@/shared/...` (or any non-renderer path) will now resolve incorrectly, as `@` no longer points to the `src` root. The change was made to satisfy the new test file's `@/stores/settings-store` import (which lives under `src/renderer/stores`), but it silently breaks all other tests that rely on `@` pointing to `src`. This is a data consistency / correctness problem for the entire test suite.

**Suggested fix:**
```
Keep `@` pointing to `resolve(__dirname, 'src')` and instead use the existing `@renderer` alias in the test file, or add a separate alias specifically for renderer stores. Alternatively, ensure all production code also uses `@` as `src/renderer` (unlikely given existing conventions).
```

#### 🟡 [DEEP-2] [MEDIUM] Race condition: `isUsingApiProfile` in useEffect dependency can cause stale closures and event handler leaks
📁 `apps/frontend/src/renderer/components/UsageIndicator.tsx:366`

The useEffect that registers IPC listeners (`onUsageUpdated`, `onAllProfilesUsageUpdated`) and fires initial fetches now depends on `[isUsingApiProfile]`. Every time the user toggles between API and OAuth profiles, the effect tears down and re-registers listeners and re-fetches data. However, the async `.then()` callbacks from `requestAllProfilesUsage` capture the `isUsingApiProfile` value at the time the effect ran. If the profile changes while the promise is in flight, the callback will use a stale `isUsingApiProfile` value and may set incorrect state. Additionally, repeated rapid profile switches will cause multiple concurrent fetches whose results may arrive out of order.

**Suggested fix:**
```
Add an `isCancelled` flag (set to `true` in the cleanup function) and check it before calling setState in the `.then()` callback. This is a standard pattern to avoid stale closure updates after effect cleanup.
```

#### 🟡 [DEEP-3] [MEDIUM] Derived state `showReauthForActiveAccount` and `showUnavailableReauth` computed outside useMemo risks stale values
📁 `apps/frontend/src/renderer/components/UsageIndicator.tsx:314`

`showReauthForActiveAccount` and `showUnavailableReauth` are computed as plain variables in the render body using both `isUsingApiProfile` and state variables (`usage`, `activeProfileNeedsReauth`). While these will re-compute on each render, the `activeProfileNeedsReauth` state itself is set inside IPC callbacks that also gate on `isUsingApiProfile`. If the profile changes between a state update and the next render, there's a brief window where `activeProfileNeedsReauth` is `true` (set during OAuth mode) but `isUsingApiProfile` has flipped to `true`, meaning `showUnavailableReauth` correctly becomes `false`. However, the `activeProfileNeedsReauth` state is never explicitly cleared when switching to API mode — it retains whatever value was last set. This means if the user switches from OAuth (with reauth=true) to API and then back to OAuth before a new fetch completes, the stale `true` could flash briefly.

**Suggested fix:**
```
Reset `activeProfileNeedsReauth` to `false` at the start of the useEffect when `isUsingApiProfile` changes, or derive it entirely from the latest fetch rather than persisting as separate state.
```

#### 🟡 [DEEP-4] [MEDIUM] `sendStatusChange` helper is defined inside a scope but `newStatus` may not be set on all code paths where it's called
📁 `apps/frontend/src/main/ipc-handlers/task/execution-handlers.ts:1218`

The `sendStatusChange` helper is defined early in the handler (line ~977), but `newStatus` is passed as a parameter. At lines 1218, 1236, and 1252, `sendStatusChange(newStatus)` is called in early-return paths. The variable `newStatus` must have been assigned before these calls. Looking at the diff context, `newStatus` appears to be computed earlier in the recovery flow. However, the helper sends the notification *before* the function returns the success response. If the renderer processes the status change notification and immediately re-fetches task data, the cache might not yet be invalidated (since `invalidateTasksCache` is not called in all these early-return paths), leading to the renderer seeing stale task data.

**Suggested fix:**
```
Call `projectStore.invalidateTasksCache(project.id)` before `sendStatusChange(newStatus)` in the early-return paths at lines 1218, 1236, and 1252, similar to how it's done at line 1105.
```

#### 🔵 [DEEP-5] [LOW] Test mock for `requestAllProfilesUsage` always returns `needsReauthentication: true` which conflates test scenarios
📁 `apps/frontend/src/renderer/components/UsageIndicator.test.tsx:100`

The default mock for `requestAllProfilesUsage` in `beforeEach` always returns `buildAllProfilesUsageResponse(true)` (needs re-auth). The test 'does not show OAuth re-auth UI in API profile mode' relies on the component ignoring this because `isUsingApiProfile` is true. However, this test doesn't verify the component correctly handles the case where the API returns `needsReauthentication: false` — it only tests the gating logic. A missing test for the negative case (OAuth mode with `needsReauthentication: false` should NOT show re-auth) reduces confidence.

**Suggested fix:**
```
Add a test case: OAuth mode (no activeProfileId) with `needsReauthentication: false` should NOT show re-auth UI, confirming both conditions must be true.
```

#### 🔵 [DEEP-6] [LOW] Test determines API vs OAuth mode solely by `activeProfileId` presence, not by profile type
📁 `apps/frontend/src/renderer/components/UsageIndicator.tsx:84`

The component determines `isUsingApiProfile = Boolean(activeApiProfileId)` where `activeApiProfileId` comes from `state.activeProfileId`. The naming `activeProfileId` in the settings store is ambiguous — it could refer to any profile type (OAuth or API). If an OAuth profile also sets `activeProfileId`, the gating logic would incorrectly treat it as an API profile and suppress re-auth UI. The test uses `'api-profile-1'` as the ID, but the component has no way to distinguish profile types from the ID alone.

**Suggested fix:**
```
Verify that `activeProfileId` in the settings store is only set for API/custom-endpoint profiles and is null/undefined for OAuth profiles. If both profile types can set this field, add explicit profile type checking rather than relying on truthiness.
```

---

PR #1888, opened 2026-02-22
fix: resolve setup wizard flashing on auth screen

### LINKED ISSUE #1882: Setup wizard fails at the beginning
### Checklist

- [x] I searched existing issues and this hasn't been reported

### Area

Frontend

### Operating System

Linux

### Version

2.7.6

### What happened?

Setup Wizard screen is flashing constantly.

### Steps to reproduce

1. Start application fresh
2. Setup Wizard Welcome: Get Started
3. Sign in with Anthtopic --> Configure Claude Authentication screen starts flashing

### Expected behavior

Be able to finish authentication

https://github.com/user-attachments/assets/8297f4fe-19e5-44cc-bc30-0839e1584c8a

### Logs / Screenshots

```shell
https://github.com/user-attachments/assets/8297f4fe-19e5-44cc-bc30-0839e1584c8a
```

AndyMik90:
## Summary
- Fixes the setup wizard screen flashing/flickering constantly when navigating to the "Configure Claude Authentication" step
- **Root cause:** `loadClaudeProfiles` in `OAuthStep.tsx` was defined as a plain `async` function, creating a new reference every render. Since it was a `useEffect` dependency, this caused an infinite re-render loop (effect fires → state update → re-render → new function ref → effect fires again)
- **Fix:** Wrapped `loadClaudeProfiles` in `useCallback` with `[]` deps to stabilize the function reference

Fixes #1882

## Test plan
- [x] All 3467 frontend unit tests pass
- [ ] Launch app fresh → Setup Wizard → "Sign in with Anthropic" → auth screen should render stably without flashing

## Summary by CodeRabbit

* **Refactor**
  * Improved profile-loading logic to stabilize behavior across renders and reduce unnecessary updates.
* **Bug Fixes / Privacy**
  * Adjusted authentication logging to avoid exposing email addresses in logs.

### DIFF
--- a/apps/frontend/src/renderer/components/onboarding/OAuthStep.tsx
+++ b/apps/frontend/src/renderer/components/onboarding/OAuthStep.tsx
@@ -79,7 +79,7 @@ export function OAuthStep({ onNext, onBack, onSkip }: OAuthStepProps) {
   );
 
   // Reusable function to load Claude profiles
-  const loadClaudeProfiles = async () => {
+  const loadClaudeProfiles = useCallback(async () => {
     setIsLoadingProfiles(true);
     setError(null);
     try {
@@ -95,7 +95,7 @@ export function OAuthStep({ onNext, onBack, onSkip }: OAuthStepProps) {
     } finally {
       setIsLoadingProfiles(false);
     }
-  };
+  }, []);
 
   // Load Claude profiles on mount
   useEffect(() => {
--- a/apps/frontend/src/renderer/components/onboarding/OAuthStep.tsx
+++ b/apps/frontend/src/renderer/components/onboarding/OAuthStep.tsx
@@ -231,7 +231,7 @@ export function OAuthStep({ onNext, onBack, onSkip }: OAuthStepProps) {
 
   // Handle auth terminal success
   const handleAuthTerminalSuccess = useCallback(async (email?: string) => {
-    console.warn('[OAuthStep] Auth success:', email);
+    console.warn('[OAuthStep] Auth success');
 
     // Close terminal immediately
     setAuthTerminal(null);


AndyMik90 (commented):
## 🤖 Auto Claude PR Review

PR #1888 reviewed: 6 findings (0 structural issues). Verdict: approve.

### Findings (6 selected of 6 total)

#### 🟡 [QLT-1] [MEDIUM] useCallback with empty deps closes over stale state/props
📁 `apps/frontend/src/renderer/components/onboarding/OAuthStep.tsx:82`

The `useCallback` wrapping `loadClaudeProfiles` uses an empty dependency array `[]`, but the function body calls `setClaudeProfiles`, `setError`, and `setIsLoadingProfiles` — which are stable setState references and therefore safe — but it also likely references other variables from the component scope (e.g., any values used inside the try/catch that aren't shown in the diff). With `[]` deps, if the function ever needs to read props or non-setState closures, it will capture stale values. While this fixes the infinite loop, the correct long-term fix would be to audit the full function body and list any non-setState dependencies, or move the async logic outside the component / into a custom hook.

**Suggested fix:**
```
Audit the full body of `loadClaudeProfiles` for any references to props or state variables. If none exist beyond setState calls, `[]` is correct. If any exist, add them to the dependency array. Consider adding an ESLint `react-hooks/exhaustive-deps` check to validate.
```

#### 🔵 [QLT-2] [LOW] Generic error handling swallows error details
📁 `apps/frontend/src/renderer/components/onboarding/OAuthStep.tsx:90`

The catch block (lines ~90-94, partially visible) sets an error state but it's unclear whether it logs the actual exception or preserves its type. If the catch simply does `setError(someGenericMessage)` without logging, debugging production issues will be difficult. This is a pre-existing issue but worth flagging since the function is being refactored.

**Suggested fix:**
```
Ensure the catch block logs the original error (e.g., `console.error('Failed to load Claude profiles:', err)`) in addition to setting user-facing error state.
```

#### 🔵 [QLT-3] [LOW] Missing test coverage noted in PR checklist
📁 `apps/frontend/src/renderer/components/onboarding/OAuthStep.tsx:99`

The test plan checkbox for manual verification ('Launch app fresh → Setup Wizard → auth screen should render stably without flashing') is unchecked. Given that this bug is a user-visible infinite re-render loop, it would be valuable to add a unit/integration test that asserts `loadClaudeProfiles` is only called once on mount (e.g., by mocking the API and asserting call count), preventing regression.

**Suggested fix:**
```
Add a test that renders <OAuthStep />, mocks the profile-loading API, and asserts it is called exactly once, not in a loop.
```

#### 🟡 [DEEP-1] [MEDIUM] useCallback with empty deps captures stale closure over state setters only - safe but potentially fragile
📁 `apps/frontend/src/renderer/components/onboarding/OAuthStep.tsx:82`

The `useCallback` with `[]` dependencies is safe for React state setters (which are stable), but the function body at lines 80-96 may reference other variables from the component scope that could become stale. Without seeing the full component, if `loadClaudeProfiles` ever needs to reference props or other state values (not just setters), the empty dependency array would cause stale closure bugs. However, based on the visible code, this appears safe for now since it only calls setters and presumably stable API functions.

**Suggested fix:**
```
Verify that the function body only references stable values (React state setters, module-level imports). If it ever needs to read props or state values, add them to the useCallback dependency array. Consider using useRef for values that need to be read without triggering re-creation.
```

#### 🟡 [DEEP-2] [MEDIUM] Missing cleanup/cancellation for async effect - potential state update on unmounted component
📁 `apps/frontend/src/renderer/components/onboarding/OAuthStep.tsx:82`

The `loadClaudeProfiles` function performs async work and then calls `setIsLoadingProfiles(false)` in a `finally` block. If the component unmounts while the async operation is in-flight (e.g., user navigates away via onBack/onSkip during loading), state updates will be attempted on an unmounted component. While React 18+ suppresses the warning, this is still a logic issue that could cause unexpected behavior - for example, if the component remounts, the stale promise resolution could interfere with a fresh load.

**Suggested fix:**
```
Add an AbortController or a cleanup flag via useEffect's cleanup function to cancel or ignore the result of the async operation when the component unmounts. For example: useEffect(() => { let cancelled = false; loadClaudeProfiles(cancelled); return () => { cancelled = true; }; }, [loadClaudeProfiles]);
```

#### 🔵 [DEEP-3] [LOW] Race condition if loadClaudeProfiles is called multiple times concurrently
📁 `apps/frontend/src/renderer/components/onboarding/OAuthStep.tsx:82`

The function is described as 'reusable' and is both called on mount (via useEffect) and presumably callable from other UI interactions. If invoked concurrently (e.g., rapid user interaction plus mount effect), two concurrent executions would both set `setIsLoadingProfiles(true)`, and the first to complete would set it to `false` while the second is still in-flight, causing incorrect loading state. The results from the first call could also briefly flash before being overwritten by the second.

**Suggested fix:**
```
Add a guard (e.g., check `isLoadingProfiles` before proceeding, or use an AbortController to cancel previous in-flight requests) to prevent concurrent executions from creating inconsistent state.
```

---

AndyMik90:

PR #1890, opened 2026-02-22
fix: sync kanban card stage with live agent phase

### LINKED ISSUE #1885: The card in Kanban shows that the task is in the planning stage, but in fact it is in the coding stage.
### Checklist

- [x] I searched existing issues and this hasn't been reported

### Area

Frontend

### Operating System

Windows

### Version

2.7.6

### What happened?

- On the Kanban board, the task status is “In progress.”
- The task card says that it is in the planning stage.
- When you open the window with the task and logs, you can see that it has already moved to the coding stage, but this status is not updated.

Image

Image

### Steps to reproduce

1. Start the task
2. Wait

### Expected behavior

The current stage in the card on the Kanban board should be displayed correctly.

### Logs / Screenshots

```shell

```

AndyMik90:
## Summary
- Fixes kanban board task cards showing stale "planning" stage when the task has actually progressed to "coding"
- **Root cause:** `emitPhaseFromState` in `task-state-manager.ts` used `sequenceNumber: Date.now()` (~1.7 trillion) for XState phase transitions. The renderer's `updateExecutionProgress` has an out-of-order guard that drops events where `incomingSeq < currentSeq`. Once the initial "planning" phase set `currentSeq` to a timestamp, ALL subsequent agent progress events (using small sequential integers: 1, 2, 3...) were permanently dropped, freezing the card on "planning"
- **Fix:** Changed `sequenceNumber: Date.now()` to `sequenceNumber: 0` so XState updates bypass the sequence guard and don't poison the sequence counter for subsequent agent events

Fixes #1885

## Test plan
- [x] All 3467 frontend unit tests pass
- [ ] Start a task → observe kanban board → card should update from "planning" to "coding" as the agent progresses

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed task state reporting so execution-progress events use stable sequencing and show 0% for non-final phases, preventing misleading intermediate progress values and improving real-time progress accuracy during task execution.

### DIFF
--- a/apps/frontend/src/main/task-state-manager.ts
+++ b/apps/frontend/src/main/task-state-manager.ts
@@ -362,7 +362,12 @@ export class TaskStateManager {
 
     const phase = XSTATE_TO_PHASE[xstateState] || 'idle';
 
-    // Emit execution progress with the phase derived from XState
+    // Emit execution progress with the phase derived from XState.
+    // IMPORTANT: Do NOT use Date.now() as sequenceNumber here — doing so would
+    // permanently block all subsequent agent execution-progress events, because
+    // the agent uses small sequential integers (1, 2, 3…) that are always less
+    // than a timestamp. Instead emit with sequenceNumber=0 so the agent's own
+    // events can continue updating phaseProgress once the phase is established.
     safeSendToRenderer(
       this.getMainWindow,
       IPC_CHANNELS.TASK_EXECUTION_PROGRESS,
@@ -372,7 +377,7 @@ export class TaskStateManager {
         phaseProgress: phase === 'complete' ? 100 : 50,
         overallProgress: phase === 'complete' ? 100 : 50,
         message: `State: ${xstateState}`,
-        sequenceNumber: Date.now()  // Use timestamp as sequence to ensure it's newer
+        sequenceNumber: 0
       },
       projectId
     );
--- a/apps/frontend/src/main/task-state-manager.ts
+++ b/apps/frontend/src/main/task-state-manager.ts
@@ -374,8 +374,8 @@ export class TaskStateManager {
       taskId,
       {
         phase,
-        phaseProgress: phase === 'complete' ? 100 : 50,
-        overallProgress: phase === 'complete' ? 100 : 50,
+        phaseProgress: phase === 'complete' ? 100 : 0,
+        overallProgress: phase === 'complete' ? 100 : 0,
         message: `State: ${xstateState}`,
         sequenceNumber: 0
       },


AndyMik90 (commented):
## 🤖 Auto Claude PR Review

PR #1890 reviewed: 8 findings (1 structural issues). Verdict: needs_review.

#### 🟠 [QLT-1] [HIGH] sequenceNumber: 0 allows out-of-order delivery in reverse direction
📁 `apps/frontend/src/main/task-state-manager.ts:380`

While changing from `Date.now()` to `0` fixes the original bug where large timestamps poisoned the sequence counter, it introduces a new edge case: if `emitPhaseFromState` is called *after* the agent has already emitted events with sequenceNumber >= 1, the XState-derived phase update (with sequenceNumber 0) will be silently dropped by the same out-of-order guard (`incomingSeq < currentSeq`). This means late XState transitions (e.g., transitioning to 'complete') may never reach the renderer if agent events have already incremented the counter. A more robust fix would use a separate channel/flag to bypass the sequence guard for XState-sourced events, or use a dedicated sequence counter that doesn't collide with the agent's counter space.

**Suggested fix:**
```
Consider adding a `bypassSequenceCheck: true` flag to XState-sourced events and updating the renderer's guard logic accordingly, or use a parallel sequence namespace (e.g., negative numbers or a separate field) so XState and agent events don't interfere with each other's ordering.
```

#### 🟡 [QLT-2] [MEDIUM] Magic number 0 used without named constant
📁 `apps/frontend/src/main/task-state-manager.ts:380`

The value `0` is used as a sentinel/bypass value for the sequence guard, but this semantic meaning is only conveyed through the comment block. A named constant (e.g., `SEQUENCE_BYPASS = 0`) would make the intent explicit and ensure consistency if this pattern is used elsewhere.

**Suggested fix:**
```
Define a constant like `const SEQUENCE_NUMBER_BYPASS = 0;` at module level and use it here: `sequenceNumber: SEQUENCE_NUMBER_BYPASS`
```

#### 🟡 [QLT-3] [MEDIUM] Hard-coded progress values 50 and 100 without constants or documentation
📁 `apps/frontend/src/main/task-state-manager.ts:377`

The phaseProgress and overallProgress values use magic numbers 50 and 100. While 100 for 'complete' is intuitive, 50 as the default for all non-complete phases is arbitrary and may not accurately represent progress. These should be named constants, and the choice of 50% as a fallback deserves documentation or a more nuanced mapping.

**Suggested fix:**
```
Extract named constants (e.g., `PHASE_IN_PROGRESS = 50`, `PHASE_COMPLETE = 100`) and consider whether different phases should report different progress percentages.
```

#### 🔵 [QLT-4] [LOW] Test plan has unchecked manual test item
📁 `apps/frontend/src/main/task-state-manager.ts:380`

The PR description's test plan includes an unchecked manual test (`[ ] Start a task → observe kanban board → card should update from 'planning' to 'coding'`). Given that this is the exact scenario the fix targets, merging without completing this manual verification is risky, especially since the `sequenceNumber: 0` approach has the edge case noted in QLT-1.

**Suggested fix:**
```
Complete the manual test before merging, or add an integration/e2e test that verifies the kanban card transitions through planning → coding → complete.
```

#### 🟠 [DEEP-1] [HIGH] sequenceNumber: 0 allows out-of-order XState phase events to overwrite newer agent events
📁 `apps/frontend/src/main/task-state-manager.ts:380`

Setting `sequenceNumber: 0` fixes the original problem where `Date.now()` poisoned the sequence counter, but it introduces a new race condition. If an agent has already emitted events with sequenceNumber >= 1, and then an XState transition fires (e.g., a delayed or re-entrant state change), the renderer will accept the XState event (seq 0 < currentSeq would be dropped by the guard... wait, actually seq 0 would be DROPPED by the guard `incomingSeq < currentSeq`). This means XState phase transitions that arrive AFTER any agent event will be silently discarded. For example, if the agent emits seq=1 (coding progress), and then XState transitions to 'complete', the completion event with seq=0 will be dropped, and the card will never show 'complete' from XState. The fix only works for the initial race (planning → coding) but breaks the terminal state transition (coding → complete) if the agent has already sent any sequenced events.

**Suggested fix:**
```
Use a dedicated mechanism for XState phase updates that bypasses the sequence guard entirely (e.g., a separate IPC channel, a 'force' flag on the payload, or use `sequenceNumber: Infinity` / `Number.MAX_SAFE_INTEGER` to guarantee acceptance). Alternatively, refactor the renderer to treat phase-level transitions differently from incremental progress updates, so they are never subject to the out-of-order guard.
```

#### 🟡 [DEEP-2] [MEDIUM] Hardcoded progress values (50/100) create misleading UI state
📁 `apps/frontend/src/main/task-state-manager.ts:378`

The `phaseProgress` and `overallProgress` are hardcoded to 50 for non-complete phases and 100 for complete. When sequenceNumber is 0 and this event IS accepted (i.e., it arrives before any agent events), it will set progress to 50% immediately. If the agent subsequently sends its own progress events (e.g., 10%, 20%, 30%), the user could see progress jump from 50% back to 10%, creating a confusing UX. With the old `Date.now()` approach, agent events were dropped so this regression wasn't visible — but with seq=0, the agent events WILL flow through, causing a non-monotonic progress bar.

**Suggested fix:**
```
Set phaseProgress and overallProgress to 0 (or a small initial value like 1) for non-complete phases, so that subsequent agent progress events only move the bar forward.
```

#### 🟡 [DEEP-3] [MEDIUM] Two independent sequencing domains (XState vs agent) sharing one sequence counter creates fundamental design fragility
📁 `apps/frontend/src/main/task-state-manager.ts:380`

The root cause described in the PR — two different event sources using incompatible sequencing schemes fed into one out-of-order guard — is only partially addressed. The fix changes one magic number to another magic number. Any future code that emits to TASK_EXECUTION_PROGRESS with a different sequencing assumption will re-introduce the same class of bug. There is no type-level or runtime enforcement that sequence numbers from different sources are compatible.

**Suggested fix:**
```
Introduce separate sequence counters per source (e.g., `xstateSeq` and `agentSeq`) in the renderer, or add a `source` field to the event payload so the renderer can maintain independent sequence tracking. Alternatively, use a monotonic counter shared across all emitters (e.g., an atomic incrementing counter in TaskStateManager).
```

#### 🔵 [DEEP-4] [LOW] Manual test plan item is unchecked — no automated integration test for the fixed scenario
📁 `apps/frontend/src/main/task-state-manager.ts:380`

The PR description shows the critical test case ('Start a task → observe kanban board → card should update from planning to coding') is unchecked. The fix changes subtle sequencing behavior that unit tests alone cannot validate (as evidenced by all 3467 unit tests passing before AND after the fix). Without an integration or e2e test covering this scenario, the bug is likely to regress.

**Suggested fix:**
```
Add an integration test that simulates an XState phase transition followed by agent progress events, asserting that both are received by the renderer in order and that progress updates are not dropped.
```

---

AndyMik90:

PR #1924, opened 2026-03-06
Fix roadmap generation phase mapping and progress heartbeat

VDT-91:
## Summary
Fix roadmap generation progress/state mismatch and improve visibility during long-running phases.

## Root cause
Backend roadmap runner emitted legacy phase names (`discovery`, `features`) while renderer state machine expects canonical names (`discovering`, `generating`).

## Fixes in this PR
- Normalize legacy roadmap phases in queue, renderer store, and persisted-progress load path.
- Add throttled progress heartbeats during streaming (`text-delta`/`tool-use`) so long steps do not appear frozen.

## Result
Roadmap progress now transitions correctly:
`analyzing -> discovering -> generating -> complete`

### DIFF
--- a/apps/desktop/src/main/agent/agent-queue.ts
+++ b/apps/desktop/src/main/agent/agent-queue.ts
@@ -81,7 +81,7 @@ export class AgentQueueManager {
   private emitter: EventEmitter;
   private debouncedPersistRoadmapProgress: (
     projectPath: string,
-    phase: string,
+    phase: RoadmapPhase,
     progress: number,
     message: string,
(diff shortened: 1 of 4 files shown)


VDT-91:
Plain-language bug summary: roadmap progress events used old phase names (discovery, eatures) while the frontend state machine only accepts (discovering, generating). That mismatch caused state errors / stuck progress behavior. This PR normalizes those phases at queue/store/load boundaries and adds heartbeat progress updates so long AI steps remain visibly active.

